PR #18, opened 2021-04-13
Change links to new repository location

errt:
Changes all links to the new URL


TorunR (approved):

TorunR:
Some things I noticed: Maybe we should remove the gitter link?

errt:
Would be reasonable, yes, but I'd leave that to the planned overhaul of the website in general.

TorunR:
Yes, but I meant the Readme.md. But on the other hand we should think about the alternatives. I guess leaving a mailto opal@st... would be best. Or we leave Gitter in and find a *consistent* way to get Email Notifications

PR #25, opened 2021-05-16
Removes ':' from the temp file path that is invalid on Windows

TorunR:
I'm not sure why this path was prefixed with "jar:" for the temp file. On windows thats not a valid symbol in a path and results in a test failure.


errt (dismissed):
Looks good to me. Did you test it?

TorunR:
Well I found it with sbt test on windows and now the tests run on both windows and the CI.

errt (approved):

PR #26, opened 2021-05-16
Update to SBT 1.5.2 and fix deprecation and other warnings

TorunR:
SBT > 1.5.0 deprecates the "in" Syntax...


TorunR:
I guess this is blocked until [this issue](https://github.com/scalate/scalate/issues/309) is fixed

errt:
What is the state of this PR now?

errt:
As far as I understand, we can't do this, because for sbt 1.5.2 we would need to go to scalate 1.9.7 which requires scala-parser-combinators 2.0.0 which currently isn't compatible with other dependencies that we have, such as the sbt-pgp plugin, right? Looks like we have to postpone this PR then until this incompatibility is resolved. Not a major issue though to not update sbt right now, is it?

TorunR:
I don't think it is a problem if we postpone this. I initially started this because of Scala 3, which has other blockers anyway

errt:
Any news on this? If possible, I would really like to upgrade to a more recent version of SBT.

TorunR:
The scalate issue is resolved, and the CI and for me locally it worked.

errt (approved):

PR #29, opened 2021-06-18
Fix JDK8DependenciesVisualization

tdurieux:
d3 was not loaded with https and therefore was not loaded in an https context


errt:
Thanks for pointing this out. I guess we should just always load it with https, that should work in an http context too.

tdurieux:
It would also work. I can change that.

errt:
Perfect, thanks

errt (approved):

PR #37, opened 2021-07-09
AndroidEntryPointFinder added

Diarilex:


TorunR (commented):
Are you planning to add support for the androidx/support library version of Activity etc? e.g. https://developer.android.com/reference/androidx/activity/package-summary or https://developer.android.com/reference/android/support/packages

OPAL/br/src/main/scala/org/opalj/br/analyses/cg/EntryPointFinder.scala
```diff
@@ -347,4 +347,79 @@ object AllEntryPointsFinder extends EntryPointFinder {
...
+ * The AndroidEntryPointFinder considers specific methods of app components as entry points.
+ *
+ * @author Tom Nikisch
+ */
+object AndroidEntryPointsFinder extends EntryPointFinder {
+    override def collectEntryPoints(project: SomeProject): Traversable[Method] = {
+        val classHierarchy = project.classHierarchy
+        val eps = ArrayBuffer.empty[Method]
+
+        var peps: List[String] = List("onCreate", "onRestart", "onStart", "onResume", "onStop", "onDestroy",
```
errt: Do not use a 'var', use individual 'val's defined outside of the method.

OPAL/br/src/main/scala/org/opalj/br/analyses/cg/EntryPointFinder.scala
```diff
@@ -347,4 +347,79 @@ object AllEntryPointsFinder extends EntryPointFinder {
...
+            "onActivityResult")
+        classHierarchy.foreachSubclass(ObjectType("android/app/Activity"), project) { sc ⇒
+            for (pep ← peps) {
+                val mc = sc.findMethod(pep)
+                for (m ← mc) {
+                    if (m.body != null) {
```
errt: m.body should never be null, it's an Option, thus either using isDefined or just mapping over it as part of the 'for' should work

OPAL/br/src/main/scala/org/opalj/br/analyses/cg/EntryPointFinder.scala
```diff
@@ -347,4 +347,79 @@ object AllEntryPointsFinder extends EntryPointFinder {
...
+object AndroidEntryPointsFinder extends EntryPointFinder {
+    override def collectEntryPoints(project: SomeProject): Traversable[Method] = {
+        val classHierarchy = project.classHierarchy
+        val eps = ArrayBuffer.empty[Method]
+
+        var peps: List[String] = List("onCreate", "onRestart", "onStart", "onResume", "onStop", "onDestroy",
+            "onActivityResult")
+        classHierarchy.foreachSubclass(ObjectType("android/app/Activity"), project) { sc ⇒
+            for (pep ← peps) {
+                val mc = sc.findMethod(pep)
```
errt: Maybe better use Scala's extended 'for' syntax to have just one 'for' that iterates over 'peps', 'mc' and possibly even 'm.body' (see next comment).

OPAL/br/src/main/scala/org/opalj/br/analyses/cg/EntryPointFinder.scala
```diff
@@ -347,4 +347,79 @@ object AllEntryPointsFinder extends EntryPointFinder {
...
+ * @author Tom Nikisch
+ */
+object AndroidEntryPointsFinder extends EntryPointFinder {
+    override def collectEntryPoints(project: SomeProject): Traversable[Method] = {
+        val classHierarchy = project.classHierarchy
+        val eps = ArrayBuffer.empty[Method]
+
+        var peps: List[String] = List("onCreate", "onRestart", "onStart", "onResume", "onStop", "onDestroy",
+            "onActivityResult")
+        classHierarchy.foreachSubclass(ObjectType("android/app/Activity"), project) { sc ⇒
```
errt: Are you aware that 'foreachSubclass' a) is not reflexive (i.e., it won't include the class file for the Activity class here) and b) may include a subclass multiple times if the given ObjectType is an interface (Activity isn't, but I didn't check your other invocations of this method)?

OPAL/br/src/main/scala/org/opalj/br/analyses/cg/EntryPointFinder.scala
```diff
@@ -347,4 +347,79 @@ object AllEntryPointsFinder extends EntryPointFinder {
...
+                val mc = sc.findMethod(pep)
+                for (m ← mc) {
+                    if (m.body != null) {
+                        eps.append(m)
+                    }
+                }
+            }
+        }
+
+        peps = List("onCreate", "onStartCommand", "onBind", "onStart")
```
errt: Code is utterly repetitive, please extract a method to do this.

OPAL/br/src/main/scala/org/opalj/br/analyses/cg/EntryPointFinder.scala
```diff
@@ -347,4 +347,79 @@ object AllEntryPointsFinder extends EntryPointFinder {
...
+                for (m ← mc) {
+                    if (m.body != null) {
+                        eps.append(m)
+                    }
+                }
+            }
+        }
+
+        peps = List("onLocationChanged", "onProviderDisabled", "onProviderEnabled", "onStatusChanged")
+        classHierarchy.foreachSubtype(ObjectType("android/location/LocationListener")) { st ⇒
```
errt: Why are you using foreachSubtype here, but foreachSubclass on the previous ones? foreachSubclass directly provides you with the classfile, already ensuring that it is available.

OPAL/br/src/main/scala/org/opalj/br/analyses/cg/EntryPointFinder.scala
```diff
@@ -355,71 +355,47 @@ object AllEntryPointsFinder extends EntryPointFinder {
  * @author Tom Nikisch
  */
 object AndroidEntryPointsFinder extends EntryPointFinder {
+
+    val ActivityEPS: List[String] = List("onCreate", "onRestart", "onStart", "onResume", "onStop", "onDestroy",
```
errt: Use camel case with lower case first letter here.

OPAL/br/src/main/scala/org/opalj/br/analyses/cg/EntryPointFinder.scala
```diff
@@ -355,71 +355,47 @@ object AllEntryPointsFinder extends EntryPointFinder {
...
+            val mc = sc.findMethod("onNmeaMessage")
+            for (m ← mc if m.body.isDefined)
+                eps.append(m)
         }
         eps
-    }
\ No newline at end of file
+    }
+
+    def verifyEPS(possibleEPS: List[String], sc: ClassFile): ArrayBuffer[Method] = {
```
errt: Can you move the foreachSubclass invokation to this method, too? You can just pass List("onNmeaMessage") for the last one (or, e.g. use Traversable[String] for the parameter type).

errt: Why is this called "verifyEPS"? What verification does it perform?

OPAL/br/src/main/scala/org/opalj/br/analyses/cg/EntryPointFinder.scala
```diff
@@ -355,71 +355,47 @@ object AllEntryPointsFinder extends EntryPointFinder {
...
+        val eps = ArrayBuffer.empty[Method]
+        for (pep ← possibleEPS; m ← sc.findMethod(pep) if m.body.isDefined && !eps.contains(m)) {
```
errt: Depending on how large eps grows, it may be more (or less) performant to use a Set (where contains is constant time instead of linear, but addition may be more costly).

OPAL/br/src/main/scala/org/opalj/br/analyses/cg/EntryPointFinder.scala
```diff
@@ -355,71 +355,47 @@ object AllEntryPointsFinder extends EntryPointFinder {
...
+            eps.append(m)
```
errt: You may want to use += instead of append (and ++= instead of appendAll).

OPAL/br/src/main/scala/org/opalj/br/analyses/cg/EntryPointFinder.scala
```diff
@@ -347,4 +347,42 @@ object AllEntryPointsFinder extends EntryPointFinder {
...
+}
+
+/**
+ * The AndroidEntryPointFinder considers specific methods of app components as entry points.
+ *
+ * @author Tom Nikisch
+ */
+object AndroidEntryPointsFinder extends EntryPointFinder {
+
+    val activityEPS: List[String] = List("android/app/Activity", "onCreate", "onRestart", "onStart", "onResume",
```
TorunR: I'm not a fan of Lists in which certain indexes have different semantics, but if @errt says it's fine I don't mind.

errt: Ah, didn't notice that. Yes, please have the list only for the methods and provide the ObjectTypes individually.

TorunR: Or maybe a map?

OPAL/br/src/main/scala/org/opalj/br/analyses/cg/EntryPointFinder.scala
```diff
@@ -347,4 +347,42 @@ object AllEntryPointsFinder extends EntryPointFinder {
...
         else project.allMethodsWithBody
     }
+}
+
+/**
+ * The AndroidEntryPointFinder considers specific methods of app components as entry points.
+ *
+ * @author Tom Nikisch
+ */
+object AndroidEntryPointsFinder extends EntryPointFinder {
```
TorunR, line 358: Since you didn't answer: 
Are you planning to add support for the androidx/support library version of Activity etc? e.g. https://developer.android.com/reference/androidx/activity/package-summary or https://developer.android.com/reference/android/support/packages
If you're not doing that, please at least add a comment that makes this clear.

Diarilex: Currently i am not planning to do that but i can probably do it later. Thank you for your suggestion.

OPAL/br/src/main/scala/org/opalj/br/analyses/cg/EntryPointFinder.scala
```diff
@@ -347,4 +347,44 @@ object AllEntryPointsFinder extends EntryPointFinder {
...
+        "onStatusChanged")
+    val onNmeaMessageListenerEPS: List[String] = List("onNmeaMessage")
+    val defaultEPS = Map("android/app/Activity" -> activityEPS, "android/app/Service" -> serviceEPS,
+        "android/content/ContentProvider" -> contentProviderEPS,
+        "android/location/LocationListener" -> locationListenerEPS,
+        "android/location/onNmeaMessageListener" -> onNmeaMessageListenerEPS)
+
+    override def collectEntryPoints(project: SomeProject): Traversable[Method] = {
+        val eps = ArrayBuffer.empty[Method]
+        for (key ← defaultEPS.keys) {
```
errt: Last change, then I guess I'll accept it:
Please use for((k,v) <- map) syntax here and use meaningful names for k and v.
You then obviously don't need to query the map for the key as you got the value already in v.

errt (changes requested):

PR #40, opened 2021-09-10
Unify call graphs

errt:
Unifies the different call graph algorithms to a single base algorithm that uses different `TypeProvider`s to resolve the possible receivers of virtual calls.
This enables all call graph modules to fully utilize the precision of the different TypeProviders, including information on allocation sites where available (the reflection module does not use allocation site information in all places possible to avoid excessive complexity, but it uses it in the most important cases).
This pull request also includes preparations to allow for future context-sensitive call graphs by stringing context information along where necessary.


DEVELOPING_OPAL/validate/src/test/scala/org/opalj/fpcf/CallGraphTests.scala
```diff
@@ -142,6 +166,8 @@ class CallGraphTests extends PropertiesTest {
     }
 
     describe("the CTA call graph analysis is executed") {
+        cgKey = FTACallGraphKey
```
TorunR: Shouldn't this be CTACallGraphKey?

OPAL/tac/src/main/scala/org/opalj/tac/fpcf/analyses/APIBasedAnalysis.scala
```diff
@@ -65,7 +66,8 @@ trait APIBasedAnalysis extends FPCFAnalysis {
                             caller.hasSingleDefinedMethod) {
                             newSeenCallers += ((caller, pc, isDirect))
 
-                            results ::= handleNewCaller(caller.asDefinedMethod, pc, isDirect)
+                            // FIXME The asInstanceOf obviously won't work once different context types are possible
```
TorunR, line 69: Do you want to keep the fixme? Just checking

OPAL/tac/src/main/scala/org/opalj/tac/fpcf/analyses/cg/CGState.scala
```diff
@@ -0,0 +1,43 @@
...
+
+import scala.collection.mutable
+
+import org.opalj.fpcf.EOptionP
+import org.opalj.br.Method
+import org.opalj.br.ReferenceType
+import org.opalj.tac.fpcf.properties.TACAI
+
+/**
+ * TODO: That state object is currently copy&paste and should be refactored.
```
TorunR: Same as the fixme before: Do you want to leave this in?

OPAL/tac/src/main/scala/org/opalj/tac/fpcf/analyses/APIBasedAnalysis.scala
errt: Yes, this Fixme will be solved in an upcoming pull request that introduces context-sensitive CGs.

DEVELOPING_OPAL/validate/src/test/scala/org/opalj/fpcf/CallGraphTests.scala
errt: Thanks, will fix immediately.

OPAL/tac/src/main/scala/org/opalj/tac/fpcf/analyses/cg/CGState.scala
errt: I guess this can be removed now, I think this PR already unifies the respective copies. Thank you.

TorunR (approved):

PR #53, opened 2022-01-25
Java17strictfp

JuliusNmn:
Added isStrict attribute to classes.
Edited behavior of isStrict attribute flag in Method. The attribute is ignored if Java >= 17.


OPAL/br/src/main/scala/org/opalj/br/Method.scala
```diff
@@ -468,6 +468,16 @@ final class Method private[br] (
 
     override def asMethod: this.type = this
 
+    /**
+     *
+     * @return wether this class is defined as strict. Starting from Java 17, this is true by default.
+     */
+    override def isStrict: Boolean =
+        if (this.classFile.version.major > bi.Java17MajorVersion)
```
errt: Should be >=

OPAL/br/src/main/scala/org/opalj/br/Method.scala
```diff
@@ -468,6 +468,16 @@ final class Method private[br] (
 
     override def asMethod: this.type = this
 
+    /**
+     *
+     * @return wether this class is defined as strict. Starting from Java 17, this is true by default.
```
errt, line 473: Why adding this here instead of just doing it directly in JVMMethod?

OPAL/br/src/main/scala/org/opalj/br/ClassFile.scala
```diff
@@ -327,6 +328,16 @@ final class ClassFile private (
 
     def isFinal: Boolean = (ACC_FINAL.mask & accessFlags) != 0
 
+    /**
+     *
+     * @return wether this class is defined as strict. Starting from Java 17, this is true by default.
+     */
+    def isStrict: Boolean =
+        if (this.version.major >= bi.Java17MajorVersion)
```
errt: I think this was also true by default in some very early versions of Java, I think we should treat that here as well.

OPAL/br/src/main/scala/org/opalj/br/ClassFile.scala
```diff
@@ -327,6 +328,16 @@ final class ClassFile private (
 
     def isFinal: Boolean = (ACC_FINAL.mask & accessFlags) != 0
 
+    /**
+     *
```
errt: If you add scaladoc, please do not only use tags (like the @return), but provide a description because that is what is primarily used for the scaladoc generation.

errt (changes requested):
I requested some small changes, but otherwise it looks good to me. How about adding a warning when trying to write out a classfile with version >= 17 and strictfp flag set?

OPAL/br/src/main/scala/org/opalj/br/Method.scala
JuliusNmn: JVMMethod does not have access to a classfile

OPAL/br/src/main/scala/org/opalj/br/ClassFile.scala
JuliusNmn: strictfp was introduced in java 1.2. In earlier versions, the float precision to use was unspecified. Would it be correct to return true here?

errt: Are you sure it was unspecified? The JVM spec says "Java 1.0 and 1.1 required strict evaluation of floating-point expressions." Do you have a reference where it says it was unspecified?

JuliusNmn:
There is now a warning message when writing out strictfp in unsupported class file versions

OPAL/br/src/main/scala/org/opalj/br/ClassFile.scala
errt: Did you check this?

OPAL/bc/src/main/scala/org/opalj/bc/Assembler.scala
```diff
@@ -752,6 +754,9 @@ object Assembler {
             writeShort(methods.size)
             methods foreach { m ⇒
                 serialize(m)
+                if  ((ACC_STRICT.mask & m.access_flags) != 0 && (classFile.major_version < 46) || (classFile.major_version > 60)) {
```
errt: There's an issue here with the parentheses. Unless I'm mistaken, this will always print an error for classfile version > 60 because the && binds stronger than the ||. I guess the parentheses right before and after the || should be removed to fix this?

errt (changes requested):

OPAL/br/src/main/scala/org/opalj/br/ClassFile.scala
JuliusNmn: The annotation actually does not exist for classes, I have removed it.

errt (dismissed):

errt:
Please apply a final `sbt scalariformFormat` to ensure that the CI passes.

errt (approved):

PR #59, opened 2022-03-01
Integration test files for call graph changes

errt:
Output of integration tests changed as an expected result of changes in the call graphs.


TorunR (approved):
I guess in this case reviewing really improves the code quality

PR #64, opened 2022-03-10
Java18framework

JuliusNmn:
This increases the latest supported java version to 18. With a possibly breaking change:
I have Removed calls to getObjectPendingFinalizationCount() in util.package.gc(), which were deprecated in java 18. The behavior of util.package.gc() has been changed to a single gc()-call. This new implementation is ok for future java versions, in which finalization will be removed. But is it ok for current versions?

### DIFF
--- a/OPAL/bi/src/main/scala/org/opalj/bi/package.scala
+++ b/OPAL/bi/src/main/scala/org/opalj/bi/package.scala
@@ -58,7 +58,7 @@ package object bi {
      * class file.
      */
     def jdkVersion(majorVersion: Int): String = {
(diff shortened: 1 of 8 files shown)


errt:
Would it be possible to implement this in a way that it executes the original code if possible (i.e., Java version < 18) and the new only if Java version >= 18?
Also, if there is no new bytecode to support, we don't need these `Framework` classes.

EDIT: Wait, the method is still there just deprecated, so this is 'only' a problem for building OPAL right now, right? Can we just use @nowarn to suppress this?

PR #65, opened 2022-03-11
java 18 compatibility

JuliusNmn:
increased maximum supported java version to 18
silenced deprecation warnings for calls to getObjectPendingFinalizationCount


JuliusNmn:
Build fails because of the following error:

```
[error] /home/github-actions/_work/opal/opal/OPAL/common/src/main/scala/org/opalj/util/package.scala:59:6: @nowarn annotation does not suppress any warnings
[265](https://github.com/opalj/opal/runs/5514314051?check_suite_focus=true#step:3:265)
[error]     @nowarn("msg=method getObjectPendingFinalizationCount in trait MemoryMXBean is deprecated")
```

The warning is only generated on SDK 18, so building fails.

errt:
You can try whether the `since` flag helps: https://www.scala-lang.org/2021/01/12/configuring-and-suppressing-warnings.html

JuliusNmn:
In the latest commit I have globally suppressed all "unused @nowarn" warnings. In my opinion, this is the best option for now until gc() is updated for Java 18.

errt:
So it is not possible to suppress this locally? In this case, I agree that this is probably the best (or least bad) option for now.

errt (approved):

PR #69, opened 2022-05-27
Throw RuntimeException if jmods are unlocatable

t1mlange:
This change throws an exception if the jmods folder doesn't exist, similar to what already happens for Java 8, and stops OPAL from silently failing to include the JRE.

E.g. Red Hat derivatives package the JDK and jmods separately.


errt (dismissed):
LGTM

errt:
Please run `sbt scalariformFormat` to get the formatting right.

errt (approved):

PR #73, opened 2022-07-18
Allow for flushing cached types when initializing multiple projects in a row

johannesduesing:
**Reason for this PR**
As described in `org.opalj.br.Type:1059` (see [here on develop](https://github.com/opalj/opal/blob/6b89b9d4b9994a8845541fe35015cf8da707f91b/OPAL/br/src/main/scala/org/opalj/br/Type.scala#L1059)), the static caches for objects of type `ObjectType` and `ArrayType` will grow continuously when creating multiple `Project` instances in a row. This is a problem when processing a large number of JAR files, as it will eventually consume all memory and cause the JVM to crash.

**Changes in this PR**
This PR introduces the two methods `ObjectType.flushTypeCache()` and `ArrayType.flushTypeCache()`, which both flush the respective caches. The methods will not flush predefined instances of `ObjectType`s and `ArrayType`s, as those are referenced via final vals. Invalidating those would negate any speedup that they originally achieved.

Obviously those methods can only be safely used when invoked after all analyses for one project are done, and before another one is created. 

This PR also introduces test case to assert that:
* All non-predefined types are in fact removed
* No predefined types are removed

This is my first contribution to OPAL, please let me know if there's anything you'd like me to change 😃

### DIFF
--- a/OPAL/br/src/main/scala/org/opalj/br/Type.scala
+++ b/OPAL/br/src/main/scala/org/opalj/br/Type.scala
@@ -10,7 +10,9 @@ import java.util.{Arrays => JArrays}
 import java.util.concurrent.atomic.AtomicInteger
 import java.util.concurrent.locks.ReentrantReadWriteLock
 
-import scala.collection.{SortedSet, mutable}
+import scala.collection.mutable
+import scala.collection.SortedSet
+
 import scala.math.Ordered
 
 import org.opalj.collection.UIDValue
@@ -1057,10 +1059,12 @@ final class ObjectType private ( // DO NOT MAKE THIS A CASE CLASS!
 object ObjectType {
 
     /**
-     * Flush the ObjectType caches (not including the predefined types to keep performance benefits of final vals)
+     *  Flushes the global cache for ObjectType instances. This does not include the predefined types,
+     *  which are kept in memory for performance reasons.
      */
     def flushTypeCache(): Unit = {
-        // First we need to write all cached new types to the actual array, otherwise we might delete the predefined types
+        // First we need to write all cached new types to the actual array, otherwise
+        // we might delete the predefined types
         updateObjectTypes()
 
         val writeLock = cacheRWLock.writeLock()
@@ -1070,20 +1074,21 @@ object ObjectType {
             // Find all keys in the Cache-Map that belong to ObjectTypes that are not predefined
             val keysToRemove = mutable.HashSet[String]()
 
-            this.cache.forEach { case (key: String, obj: WeakReference[ObjectType]) =>
-                if( obj.get().id > HighestPredefinedTypeId ){
-                    keysToRemove.add(key)
-                }
+            this.cache.forEach {
+                case (key: String, obj: WeakReference[ObjectType]) =>
+                    if (obj.get().id > HighestPredefinedTypeId) {
+                        keysToRemove.add(key)
+                    }
             }
 
             // Remove all ObjectTypes that are not predefined
             keysToRemove.foreach { cache.remove }
 
-            // Truncate the ObjectType cache array to loose all not-predefined ObjectTypes are removed
-           this.objectTypes = JArrays.copyOf( this.objectTypes, HighestPredefinedTypeId)
+            // Truncate the ObjectType cache array to loose all not-predefined ObjectTypes
+            this.objectTypes = JArrays.copyOf(this.objectTypes, HighestPredefinedTypeId)
 
             // Reset ID counter to highest id in the cache
-            this.nextId.set( HighestPredefinedTypeId + 1 )
+            this.nextId.set(HighestPredefinedTypeId + 1)
         } finally {
             writeLock.unlock()
         }
@@ -1555,25 +1560,35 @@ final class ArrayType private ( // DO NOT MAKE THIS A CASE CLASS!
  */
 object ArrayType {
 
+    /**
+     *  Flushes the global cache for ArrayType instances. This does not include the predefined types,
+     *  which are kept in memory for performance reasons.
+     */
     def flushTypeCache(): Unit = {
-        // First we need to write all cached new types to the actual array, otherwise we might delete the predefined types
+        // First we need to write all cached new types to the actual array, otherwise
+        // we might delete the predefined types
         updateArrayTypes()
 
         cache.synchronized {
 
+            // Collect all keys of ATs that are not predefined
             val keysToRemove = mutable.HashSet[FieldType]()
 
-            this.cache.forEach { case (compT: FieldType, refAT: WeakReference[ArrayType]) =>
-                if( refAT.get().id < LowestPredefinedTypeId ){
-                    keysToRemove.add(compT)
-                }
+            this.cache.forEach {
+                case (compT: FieldType, refAT: WeakReference[ArrayType]) =>
+                    if (refAT.get().id < LowestPredefinedTypeId) {
+                        keysToRemove.add(compT)
+                    }
             }
 
+            // Remove all non-predefined ATs from cache
             keysToRemove.foreach { cache.remove }
 
-            this.arrayTypes = JArrays.copyOf( this.arrayTypes , -LowestPredefinedTypeId + 1)
+            // Reset array to only contain predefined ATs
+            this.arrayTypes = JArrays.copyOf(this.arrayTypes, -LowestPredefinedTypeId + 1)
 
-            this.nextId.set( LowestPredefinedTypeId - 1 )
+            // Reset id counter
+            this.nextId.set(LowestPredefinedTypeId - 1)
 
         }
     }
--- a/OPAL/br/src/main/scala/org/opalj/br/Type.scala
+++ b/OPAL/br/src/main/scala/org/opalj/br/Type.scala
@@ -3,16 +3,13 @@ package org.opalj
 package br
 
 import scala.annotation.tailrec
-
 import java.lang.ref.WeakReference
 import java.util.WeakHashMap
 import java.util.{Arrays => JArrays}
 import java.util.concurrent.atomic.AtomicInteger
 import java.util.concurrent.locks.ReentrantReadWriteLock
-
-import scala.collection.SortedSet
+import scala.collection.{SortedSet, mutable}
 import scala.math.Ordered
-
 import org.opalj.collection.UIDValue
 import org.opalj.collection.immutable.UIDSet
 import org.opalj.collection.immutable.UIDSet2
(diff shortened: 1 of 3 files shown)


OPAL/br/src/main/scala/org/opalj/br/Type.scala
```diff
@@ -1056,6 +1058,43 @@ final class ObjectType private ( // DO NOT MAKE THIS A CASE CLASS!
...
+                case (key: String, obj: WeakReference[ObjectType]) =>
+                    if (obj.get().id > HighestPredefinedTypeId) {
+                        keysToRemove.add(key)
+                    }
+            }
+
+            // Remove all ObjectTypes that are not predefined
+            keysToRemove.foreach { cache.remove }
+
+            // Truncate the ObjectType cache array to loose all not-predefined ObjectTypes
```
errt: Typo: loose -> lose

OPAL/br/src/main/scala/org/opalj/br/Type.scala
```diff
@@ -1056,6 +1058,43 @@ final class ObjectType private ( // DO NOT MAKE THIS A CASE CLASS!
...
+        updateObjectTypes()
+
+        val writeLock = cacheRWLock.writeLock()
+
+        writeLock.lock()
+        try {
+            // Find all keys in the Cache-Map that belong to ObjectTypes that are not predefined
+            val keysToRemove = mutable.HashSet[String]()
+
+            this.cache.forEach {
```
errt: Remove 'this' (everywhere where it's unnecessary, not just here)

OPAL/br/src/main/scala/org/opalj/br/Type.scala
```diff
@@ -1333,6 +1372,8 @@ object ObjectType {
     final val ObjectInputStream = ObjectType("java/io/ObjectInputStream")
     final val ObjectOutputStream = ObjectType("java/io/ObjectOutputStream")
 
+    private[br] final val HighestPredefinedTypeId = nextId.get() - 1
```
errt: I'd prefer to start these with a lower case letter.

OPAL/br/src/main/scala/org/opalj/br/Type.scala
```diff
@@ -1519,6 +1560,39 @@ final class ArrayType private ( // DO NOT MAKE THIS A CASE CLASS!
  */
 object ArrayType {
 
+    /**
+     *  Flushes the global cache for ArrayType instances. This does not include the predefined types,
+     *  which are kept in memory for performance reasons.
+     */
+    def flushTypeCache(): Unit = {
```
errt: Should we have a method (probably on ReferenceType) that invokes both this and the ObjectType cache flush to do both at once?

OPAL/br/src/test/scala/org/opalj/br/TypeTest.scala
```diff
@@ -20,4 +20,89 @@ class TypeTest extends AnyFunSpec {
...
+        }
+
+        it("should not remove predefined types on flush") {
+            ObjectType.flushTypeCache()
+
+            // Check that first and last predefined OT are still present after flush
+            val obj = ObjectType.lookup(ObjectType.ObjectId)
+            assert(obj.fqn.equals("java/lang/Object"))
+
+            val oop = ObjectType.lookup(ObjectType.HighestPredefinedTypeId)
```
errt: Should probably be 'oos'?

OPAL/br/src/main/scala/org/opalj/br/Type.scala
```diff
@@ -1056,6 +1058,43 @@ final class ObjectType private ( // DO NOT MAKE THIS A CASE CLASS!
...
+        // First we need to write all cached new types to the actual array, otherwise
+        // we might delete the predefined types
+        updateObjectTypes()
+
+        val writeLock = cacheRWLock.writeLock()
+
+        writeLock.lock()
+        try {
+            // Find all keys in the Cache-Map that belong to ObjectTypes that are not predefined
+            val keysToRemove = mutable.HashSet[String]()
```
errt: Why not just mutable.Set?

errt: Actually, instead of collecting the keys, couldn't you just traverse `objectTypes` starting from HighestPredefinedTypeId and delete these?

OPAL/br/src/main/scala/org/opalj/br/Type.scala
```diff
@@ -10,7 +10,9 @@ import java.util.{Arrays => JArrays}
 import java.util.concurrent.atomic.AtomicInteger
 import java.util.concurrent.locks.ReentrantReadWriteLock
 
+import scala.collection.mutable
 import scala.collection.SortedSet
+
```
errt: I don't think there should be a blank line here.

OPAL/br/src/main/scala/org/opalj/br/Type.scala
```diff
@@ -1056,6 +1058,43 @@ final class ObjectType private ( // DO NOT MAKE THIS A CASE CLASS!
  */
 object ObjectType {
 
+    /**
+     *  Flushes the global cache for ObjectType instances. This does not include the predefined types,
+     *  which are kept in memory for performance reasons.
+     */
+    def flushTypeCache(): Unit = {
```
errt: Please move this after the lookup method.

errt (changes requested):
Looks good to me, apart from a few minor cosmetic issues. Thanks for contributing!

OPAL/br/src/main/scala/org/opalj/br/Type.scala
johannesduesing: Yes that's a good idea. I was thinking about where to this method, but ReferenceType makes perfect sense to me. Will do right away 👍

OPAL/br/src/main/scala/org/opalj/br/Type.scala
johannesduesing: Yes that could also work... Just one doubt i have in mind, maybe you could comment on that:
There could be very rare thread interleavings where after `updateObjectTypes()` (which synchronizes `cache` and `objectTypes`) new OTs are created and added to `cache` before `flushTypeCache` acquires the lock. In those cases, this method you are proposing would not clear those new OTs from the cache (since they are not yet contained in `objectTypes`). This would create a problem, as the ID would be reset while there are still OTs with a higher ID in the `cache`, which would "materialize" into the `objectTypes` array on the next call to `updateObjectTypes()`. Do you think that's an actual problem?

errt: Agreed, that should be prevented. I think it should work if you move the writeLock before invoking updateObjectTypes?

johannesduesing: Yep that works, i just implemented it the way you suggested.

johannesduesing:
I just noticed that the usage of `flushTypeCache` in the testing process actuall breaks a test that (to me) looks completely unrelated: The `ClassFileTest` now fails for five test cases. I'll investigate that and get back to you...

johannesduesing:
I think i've identified the problem. The `ClassFileTest` defines a bunch of `val`s that include instances of `Project` and also `ObjectType`. These values are initialized when the class is loaded, so before any test is actually run.  Then some tests are executed, which includes the `TypeTest` that executes `flushTypeCache()`. Then when the testcases in `ClassFileTest` are executed, they try to compare (via `eq`) OTs that where created before the flush (when the `val`s were initialized) with OTs that are created inside the testcase, so after the flush. Since comparison of OTs is ID-based, this comparison is now broken. See an example here:

```
  val innerclassesProject = TestSupport.biProject("innerclasses-1.8-g-parameters-genericsignature.jar")
  val outerClass = innerclassesProject.classFile(ObjectType("innerclasses/MyRootClass")).get
  val innerPrinterOfXClass = innerclassesProject.classFile(ObjectType("innerclasses/MyRootClass$InnerPrinterOfX")).get
  val formatterClass = innerclassesProject.classFile(ObjectType("innerclasses/MyRootClass$Formatter")).get

  test("that all direct nested classes of a top-level class are correctly identified") {
      outerClass.nestedClasses(innerclassesProject).toSet should be(Set(
          ObjectType("innerclasses/MyRootClass$1"),
          ObjectType("innerclasses/MyRootClass$1MyInnerPrinter"),
          ObjectType("innerclasses/MyRootClass$2"),
          ObjectType("innerclasses/MyRootClass$Formatter"),
          ObjectType("innerclasses/MyRootClass$InnerPrinterOfX")
      ))
  }
```

I certainly don't want to impose an implicit ordering on testcase execution. My first idea for a fix would be to convert the `val`s in `ClassFileTest` to `lazy val`s, which ensures all OTs are created after the flush, and all comparisons are valid. Does that sound reasonable to you @errt ?

errt:
Could this maybe be solved by changing ClassFileTest from FunSuite to FunSpec and move the vals into a `describe`? I think that would prevent them from being initialized early.

On another note: please run `sbt scalariformFormat` on your PR, as currently the CI fails for formatting issues.

johannesduesing:
I just tried that locally, unfortunately it exhibits the exact same behavior as with `FunSuite` when i execute all tests for BR. Here's a snippet of what i tried:

```
class ClassFileTest extends AnyFunSpec with Matchers {
  describe("ClassFile") {

        val innerclassesProject = TestSupport.biProject("innerclasses-1.8-g-parameters-genericsignature.jar")
        val outerClass = innerclassesProject.classFile(ObjectType("innerclasses/MyRootClass")).get
        val innerPrinterOfXClass = innerclassesProject.classFile(ObjectType("innerclasses/MyRootClass$InnerPrinterOfX")).get
        val formatterClass = innerclassesProject.classFile(ObjectType("innerclasses/MyRootClass$Formatter")).get

        it("should correctly identify all direct nested classes of a top-level class"){
            outerClass.nestedClasses(innerclassesProject).toSet should be(Set(
                ObjectType("innerclasses/MyRootClass$1"),
                ObjectType("innerclasses/MyRootClass$1MyInnerPrinter"),
                ObjectType("innerclasses/MyRootClass$2"),
                ObjectType("innerclasses/MyRootClass$Formatter"),
                ObjectType("innerclasses/MyRootClass$InnerPrinterOfX")
            ))
        }
  }
}
```

This still seems to be executed before all test cases, or at least before the testcase that flushes the cache. Adding the `lazy` keyword seems to solve this issue in both scenarios (at least on my machine).

errt:
Ok, then let's just go with the `lazy vals`.

errt (approved):
Perfect, thanks.

errt:
Unfortunately, this seems to have introduced a non-deterministic error:
https://github.com/opalj/opal/runs/7409513069?check_suite_focus=true
(the error is around line 7900)
Can you please check? It might have to do with the weak refererences in the cache. Maybe we have to use your original solution of getting the keys to remove from the cache instead of the array, or at least guard against NPEs.

johannesduesing:
That's unfortunate... I'll look into it tomorrow and try to find a solution. My first idea right now would be to just completely clear the cache and re-fill it after the array has been truncated. Since the method now holds a lock / monitor the entire time, this should also be possible. But i'll try to reproduce the error first and get back to you.

errt:
Yes, that's also a solution that I thought about before. This should be easy to do, as the pre-defined types are still available in the array to add them back into the cache.

PR #75, opened 2022-07-20
Fix inconsistencies in test execution when flushing type caches

johannesduesing:
**Reason for this PR**
#73 has introduced `ObjectType.flushTypeCache()`, `ArrayType.flushTypeCache()` and corresponding test cases. As mentioned [here](https://github.com/opalj/opal/pull/73#issuecomment-1190151306), these tests seem to (randomly) induce test failures.

**Changes in this PR**
As mentioned in [my reply](https://github.com/opalj/opal/pull/73#issuecomment-1190224681) to the original comment, this PR  changes the algorithm for flushing the caches. Now the `cache` is completely cleared, and re-filled afterwards.

I want to point out that i ran the `TypeTest`, the entire BR test suite and all OPAL unit tests multiple times, and did not encounter any error. However, i was also unable to reproduce the initial error on my machine, so i can't guarantee that this fixes the issue.

### DIFF
--- a/OPAL/br/src/main/scala/org/opalj/br/Type.scala
+++ b/OPAL/br/src/main/scala/org/opalj/br/Type.scala
@@ -1125,12 +1125,16 @@ object ObjectType {
         updateObjectTypes()
 
         try {
-            // Remove all non-predefined OTs from the cache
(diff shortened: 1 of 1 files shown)


errt (approved):
LGTM

PR #96, opened 2022-12-22
Partly Bug Fix Issue 87

roterEmil:


OPAL/ai/src/main/scala/org/opalj/ai/domain/l1/ClassValues.scala
```diff
@@ -246,10 +248,24 @@ private object ClassValues {
 
     final val forName_String = MethodDescriptor(ObjectType.String, ObjectType.Class)
 
+    final val forName_String_Caller = {
```
errt, line 251: These should be _Class, the last parameter is a Class value after all

errt (changes requested):
As an aside: I know we're short on time today, but I'd value proper pull request titles and descriptions, even if short. They help documenting for any future changes.

PR #99, opened 2023-01-12
using sorted sets for dependent immutability to pass integration tests

roterEmil:


OPAL/br/src/main/scala/org/opalj/br/fpcf/properties/immutability/FieldImmutability.scala
```diff
@@ -65,7 +67,7 @@ case class DependentlyImmutableField(parameter: Set[String]) extends FieldImmuta
             that
         else {
             if (that.isInstanceOf[DependentlyImmutableField])
```
errt: 1. Don't nest an `if` inside an `else`, use `else if`
2. More importantly: Don't use instanceof, use pattern matching! (In this case, for the whole method)

OPAL/br/src/main/scala/org/opalj/br/fpcf/properties/immutability/ClassImmutability.scala
```diff
@@ -55,8 +60,9 @@ case object TransitivelyImmutableClass extends ClassImmutability {
     def meet(that: ClassImmutability): ClassImmutability = that
 }
 
-case class DependentlyImmutableClass(parameter: Set[String]) extends ClassImmutability {
+case class DependentlyImmutableClass(parameter: SortedSet[String]) extends ClassImmutability {
```
errt: There's so many lines where the parameter is initialized with SortedSet.empty, maybe it would be beneficial to put this as a default?

OPAL/br/src/main/scala/org/opalj/br/fpcf/properties/immutability/ClassImmutability.scala
```diff
@@ -55,8 +60,9 @@ case object TransitivelyImmutableClass extends ClassImmutability {
...
 
+    override def toString: String = super.toString
```
errt: Why?

errt (changes requested):

OPAL/br/src/main/scala/org/opalj/br/fpcf/properties/immutability/ClassImmutability.scala
errt: Then on the other hand, this is only in the matchers, so we might not want to pollute the code here with something that has no meaning elsewhere. Why do the matchers even have that many DependentlyImmutables with empty sets?

OPAL/br/src/main/scala/org/opalj/br/fpcf/properties/immutability/FieldImmutability.scala
```diff
@@ -62,15 +62,12 @@ case class DependentlyImmutableField(parameter: SortedSet[String]) extends Field
...
-        else {
-            if (that.isInstanceOf[DependentlyImmutableField])
-                DependentlyImmutableField(parameter ++ that.asInstanceOf[DependentlyImmutableField].parameter)
-            else this
-        }
-    }
+    def meet(that: FieldImmutability): FieldImmutability = that match {
+        case MutableField | NonTransitivelyImmutableField => that
+        case DependentlyImmutableField(_) =>
+          DependentlyImmutableField(parameter ++ that.asInstanceOf[DependentlyImmutableField].parameter)
```
errt: Now finally remove that asInstanceOf. That was the whole point of the pattern matching...

errt (changes requested):

OPAL/br/src/main/scala/org/opalj/br/fpcf/properties/immutability/FieldImmutability.scala
errt: Also, didn't we agree to rename this to `parameters`?

errt (approved):

PR #100, opened 2023-01-17
fixed non determinism issue in field immutability

roterEmil:


OPAL/tac/src/main/scala/org/opalj/tac/fpcf/analyses/FieldImmutabilityAnalysis.scala
```diff
@@ -158,18 +157,8 @@ class L0FieldImmutabilityAnalysis private[analyses] (val project: SomeProject)
                                 case _                          => innerPackageIdentifier
                             }
                             val objectType = ObjectType(objectPath)
-                            propertyStore(objectType, TypeImmutability.key) match {
-
-                                case LBP(TransitivelyImmutableType) => // transitively immutable is default
-
-                                //nested generic classes are over-approximated as mutable
-                                case UBP(DependentlyImmutableType(_) | NonTransitivelyImmutableType | MutableType) =>
```
errt, line 166: What happened to this DependentlyImmutableType case? Is it not necessary anymore?

OPAL/tac/src/main/scala/org/opalj/tac/fpcf/analyses/FieldImmutabilityAnalysis.scala
```diff
@@ -203,58 +192,66 @@ class L0FieldImmutabilityAnalysis private[analyses] (val project: SomeProject)
...
-                                            DependentlyImmutableField(state.genericTypeParameters).meet(state.upperBound)
-                                    case _ =>
-                                }
-                            case _ =>
-                        }
-                        if (!state.upperBound.isInstanceOf[DependentlyImmutableField])
-                            state.upperBound = NonTransitivelyImmutableField
-                    }
-                    case UBP(MutableClass | NonTransitivelyImmutableClass) =>
+                val result: EOptionP[ReferenceType, ClassImmutability] = propertyStore(referenceType, ClassImmutability.key)
```
errt: As you don't use result for anything else, maybe just get rid of it and merge this with the next line?

OPAL/tac/src/main/scala/org/opalj/tac/fpcf/analyses/FieldImmutabilityAnalysis.scala
```diff
@@ -86,7 +85,7 @@ class L0FieldImmutabilityAnalysis private[analyses] (val project: SomeProject)
             var fieldImmutabilityDependees: Set[EOptionP[Entity, Property]] = Set.empty,
             var fieldIsNotAssignable:       Option[Boolean]                 = None,
             var genericTypeParameters:      SortedSet[String]               = SortedSet.empty,
-            var innerType:                  Option[ObjectType]              = None,
+            var innerTypes:                 Set[ReferenceType]              = Set.empty,
```
errt, line 87: Why did this change? Can there be multiple now while there could only be one before? Why? What do multiple inner types mean compared to a single one?

OPAL/tac/src/main/scala/org/opalj/tac/fpcf/analyses/FieldImmutabilityAnalysis.scala
roterEmil: That is handled in checkTypeImmutability

roterEmil: case FinalEP(t, DependentlyImmutableType(_)) =>
                if (t != state.field.fieldType || state.innerTypes.contains(t.asReferenceType))
                    state.upperBound = NonTransitivelyImmutableField

errt: But the old version here checked for a UBP, the new one only for FinalEP, so that at least seems different to me.

errt (approved):
LGTM

PR #109, opened 2023-02-20
SQL Taint Analysis

Imagee:
Taint analysis of SQL built on IFDS.


errt:
Closed, as we will not include direct cross language analysis in OPAL in the near future.

PR #113, opened 2023-04-27
Analysis auto config

errt:
If no analysis is specified for a property, OPAL can now automatically select a default analysis.
Defaults to off to ensure backwards compatibility and because using cheap fallbacks may often still be preferential.


TorunR (approved):

johannesduesing (approved):
Looks good to me 👍

PR #120, opened 2023-06-30
Replace deprecated constructor

TorunR:
It could be the case that moving to URI instead of URL is the correct move here


errt:
Moving to URI would ripple through hundreds of classes (this is, e.g., the source of the usual Project[URL] type). I don't think it's worth it, and there might be places where the URL functionality is needed in some way.

errt (dismissed):
LGTM

errt (approved):

PR #121, opened 2023-07-05
Use SystemPropertiesAnalysis to get more statically known Strings

errt:
We have an analysis that identifies Strings set as system properties, use it for, e.g., reflection analysis.

This is WIP, as dependency handling and type changes (Set[String] instead of String) still have to be done. I won't continue on this for now.

@johannesduesing I assigned you to this, as it is part of the greater task of more proper string analysis that also includes PR #1


johannesduesing:
/it:test

*EDIT:* This comment was used to test whether the changes introduced in #155 actually work as intended - which they seem to do, the job status is set to `pending` as expected.

maximilianruesch:
Note: Work on the string analysis from PR #1 includes work on the system properties analysis. Thus, if #1 is merged, we can close this PR.

errt:
I don't think that is true. #1 works on the analysis of SystemProperties, but does not use the results for reflection analysis, which is what this PR is about. Instead #214 integrates StringAnalysis results with the reflection analysis. However, I'm not sure whether #214 would also result in the use of results from SystemProperties (I think it does) and also it only seems to use StringAnalysis results for Class.forName, not in other places that reflection analysis relies on Strings.
The work started here should probably continue in #214, but I will leave this PR open for now to remind us that #214 might need further extension.

errt:
I have updated #214 to use the string analysis for all strings used in reflection, rendering this PR obsolete.

PR #122, opened 2023-07-07
Feature/remove scalariform

TorunR:


DEVELOPING_OPAL/demos/src/main/scala/org/opalj/tac/CloseResources.scala
```diff
@@ -6,9 +6,6 @@ import java.net.URL
 
 import scala.jdk.CollectionConverters._
 
-// import org.opalj.log.OPALLogger.info
```
errt: Can we avoid the removal of commented out imports? They are sometimes useful to restore necessary imports for commented out code.

shippable.yml
errt, line 1: How is the deletion of this file related to this PR?

TorunR: Sorry, wanted to do that separately. I'll revert this.

DEVELOPING_OPAL/demos/src/main/scala/org/opalj/tac/CloseResources.scala
TorunR: I checked. Scalafix removes all comments in the imports. If we put the commented out imports below all others, they are kept.

TorunR: I think I've fixed all of these. But there might be other comments we are losing ...

DEVELOPING_OPAL/demos/src/main/scala/org/opalj/ai/ExceptionUsage.scala
```diff
@@ -23,23 +24,22 @@ object ExceptionUsage extends ProjectAnalysisApplication {
     override def doAnalyze(
         theProject:    Project[URL],
         parameters:    Seq[String],
-        isInterrupted: () => Boolean
-    ): BasicReport = {
+        isInterrupted: () => Boolean): BasicReport = {
```
errt, line 27: This is really ugly, can we avoid it?

DEVELOPING_OPAL/demos/src/main/scala/org/opalj/ai/ExceptionUsage.scala
```diff
@@ -52,24 +52,22 @@ object ExceptionUsage extends ProjectAnalysisApplication {
             //    of all exceptions (objects that are an instance of Throwable)
             //    (Since exceptions are never created in a register, it is sufficient
             //    to analyze the operands.)
-            def collectExceptions(value: result.domain.DomainSingleOriginReferenceValue): Unit = {
+            def collectExceptions(value: result.domain.DomainSingleOriginReferenceValue): Unit =
```
errt, line 55: I'm all for removing unnecessary brackets, but I think curlies around multi-line statements help with code comprehensions. Is it possible to keep them?

DEVELOPING_OPAL/demos/src/main/scala/org/opalj/ai/ExceptionUsage.scala
```diff
@@ -78,64 +76,53 @@ object ExceptionUsage extends ProjectAnalysisApplication {
...
+                        case _                   =>
                         /*nothing to do*/
                     }
                 }
             }
 
             val usages =
-                for { ((pc, typeName), exceptionUsage) <- exceptionUsages }
-                    yield ExceptionUsage(method, pc, typeName, exceptionUsage)
+                for { ((pc, typeName), exceptionUsage) <- exceptionUsages } yield ExceptionUsage(method,
```
errt, line 113: This doesn't look nice, keeping the constructor/method call together would be good if possible.

.scalafmt.conf
```diff
@@ -1,45 +1,140 @@
...
+project.excludePaths = [
+  "glob:**/DEVELOPING_OPAL/demos/src/main/scala/org/opalj/fpcf/analyses/InterProceduralEscapeAnalysisDemo.scala",
+  "glob:**/DEVELOPING_OPAL/demos/src/main/scala/org/opalj/fpcf/analyses/SimpleEscapeAnalysisDemo.scala"
 ]
 
-literals.long = Upper
-literals.float = Lower
-literals.double = Lower
-lineEndings = unix
+fileOverride {
```
errt: What is the point of this? At least document if necessary.

.scalafmt.conf
```diff
@@ -1,45 +1,140 @@
...
+docstrings {
+  style = Asterisk
+  forceBlankLineBefore = false
+  wrap = false
+  oneline = keep
+}
 
-continuationIndent = 4
+danglingParentheses {
+ctrlSite = false
```
errt: Indent

DEVELOPING_OPAL/demos/src/main/scala/org/opalj/ai/ExceptionUsage.scala
```diff
@@ -78,64 +76,53 @@ object ExceptionUsage extends ProjectAnalysisApplication {
...
-                None
+            if (usages.isEmpty) None
             else {
                 Some(usages)
             }
         }).filter(_.isDefined).map(_.get).flatten
 
         val (notUsed, used) = usages.toSeq.partition(_.usageInformation.isEmpty)
-        var report = used.map(_.toString).toList.sorted.mkString("\nUsed\n", "\n", "\n")
+        var report          = used.map(_.toString).toList.sorted.mkString("\nUsed\n", "\n", "\n")
```
errt, line 125: Why does it align regular local variable definitions? Do we really want this?

DEVELOPING_OPAL/demos/src/main/scala/org/opalj/ai/MethodsThatAlwaysReturnAPassedParameter.scala
```diff
@@ -22,61 +23,54 @@ object MethodsThatAlwaysReturnAPassedParameter extends ProjectAnalysisApplicatio
...
         val methods = for {
             classFile <- theProject.allClassFiles.par
-            method <- classFile.methods
+            method    <- classFile.methods
             if method.body.isDefined
             if method.descriptor.returnType.isReferenceType
             if (
                 method.descriptor.parametersCount > 0 &&
-                method.descriptor.parameterTypes.exists(_.isReferenceType)
+                    method.descriptor.parameterTypes.exists(_.isReferenceType)
```
errt, line 41: I don't mind this too much, but I could live without this additional indentation.

DEVELOPING_OPAL/demos/src/main/scala/org/opalj/ai/UselessComputations.scala
```diff
@@ -24,79 +25,72 @@ object UselessComputations extends ProjectAnalysisApplication {
...
             } yield {
                 import result._
                 val results = collectPCWithOperands(domain)(body, operandsArray) {
                     case (
-                        pc,
-                        IFNULL(_) | IFNONNULL(_),
-                        Seq(domain.DomainReferenceValueTag(value), _*)
-                        ) if value.isNull.isYesOrNo =>
-                        UselessComputation(method, pc, "useless comparison with null")
+                            pc,
```
errt, line 67: Why is this indented so much?

DEVELOPING_OPAL/demos/src/main/scala/org/opalj/ai/domain/l1/IfNullParameterAnalysis.scala
```diff
@@ -45,107 +46,99 @@ import scala.collection.parallel.CollectionConverters.ImmutableIterableIsParalle
...
-                    null
-                else if (!value.isInstanceOf[domain.SingleOriginReferenceValue]) {
-                    // we are not concerned about primitive values
-                    value
-                } else {
-                    val initialReferenceValue = value.asInstanceOf[domain.SingleOriginReferenceValue]
-                    // create a new reference value which is not null
-                    initialReferenceValue.update(isNull = No)
-                }
+        def setToNonNull(domain: DefaultDomain[URL])(locals: domain.Locals): domain.Locals = locals.map { value =>
```
errt, line 60: Not very nice. Can we avoid this?

DEVELOPING_OPAL/demos/src/main/scala/org/opalj/ai/domain/l1/Ownership.scala
```diff
@@ -29,91 +30,87 @@ import scala.collection.parallel.CollectionConverters.ImmutableIterableIsParalle
...
-                                        checkValue(sorv)
-                                    case domain.DomainMultipleReferenceValuesTag(morv) =>
-                                        morv.values.forall(checkValue(_))
-                                }
-                            }
-                    }
-                    !isOwner
-                } => method
-            }
+                                            case method if {
```
errt, line 53: That's a lot of indentation...

errt (changes requested):
I think it would be good if we could break down the reformatting into different pull requests. Some uncontroversial steps that should be possible with Scalariform could be done first, to keep the load of the full scalafmt pull request lower. I'm mainly thinking about the import reordering and the string concatenation, but there may be more (e.g., maybe alignment of the arrows in for loops).

DEVELOPING_OPAL/demos/src/main/scala/org/opalj/ai/ExceptionUsage.scala
errt: I think this is the result of the "danglingParentheses" options being set to false, I guess they should be set to true, or is there a reason why they were set to false?

.scalafmt.conf
```diff
@@ -1,45 +1,140 @@
...
+optIn.breaksInsideChains = false
+rewrite {
+  rules = [
+    RedundantParens,
+    RedundantBraces,
+    SortModifiers,
+    PreferCurlyFors,
+    Imports
+  ]
+  sortModifiers.order = [
```
errt, line 86: How did you come up with this order? I think we usually have the visibility modifiers before things like final and override, so is there a reason not to keep it that way?

DEVELOPING_OPAL/demos/src/main/scala/org/opalj/ai/ExceptionUsage.scala
TorunR: There was some other stuff that was formatted with that set to true, but I don't remember what exactly happened. We'd have to try

DEVELOPING_OPAL/demos/src/main/scala/org/opalj/ai/ExceptionUsage.scala
TorunR: [This Rule and the next one I think](https://scalameta.org/scalafmt/docs/configuration.html#redundantbraces-defnbodies)

.scalafmt.conf
TorunR: We can remove that since we exclude the files explicitly now

DEVELOPING_OPAL/demos/src/main/scala/org/opalj/ai/ExceptionUsage.scala
TorunR: we have to try with `align.preset = some` and we what the impact of that is

DEVELOPING_OPAL/demos/src/main/scala/org/opalj/ai/MethodsThatAlwaysReturnAPassedParameter.scala
TorunR: I guess this can be changed with [this](https://scalameta.org/scalafmt/docs/configuration.html#indentoperator)

DEVELOPING_OPAL/demos/src/main/scala/org/opalj/ai/UselessComputations.scala
TorunR: I guess this affected by `ctorSite = 8 # Double Indent` which I just copied over.

DEVELOPING_OPAL/demos/src/main/scala/org/opalj/ai/domain/l1/IfNullParameterAnalysis.scala
TorunR: I don't know. Maybe if we change the redundant brace rules?

.scalafmt.conf
TorunR: I copied from somewhere, maybe the intellij settings? We can just change it back

errt:
> I think it would be good if we could break down the reformatting into different pull requests. Some uncontroversial steps that should be possible with Scalariform could be done first, to keep the load of the full scalafmt pull request lower. I'm mainly thinking about the import reordering and the string concatenation, but there may be more (e.g., maybe alignment of the arrows in for loops).

I'm working on something like this already, but might take a while

.scalafmt.conf
errt: Actually, override might more often be already in the front, so maybe that's something to keep. Should check later.

errt: Anyway, we should come up with one reasonable order, then apply this everywhere and stay with it. It doesn't matter too much, in any case it's better to be consistent, which currently it doesn't seem to be.

errt:
I suggest adding this:

```
literals {
  hexPrefix = Lower
  hexDigits = Upper
}
```

Could also decide on a value for `scientific`. I don't think we use scientific notation anywhere yet, but it wouldn't harm to select one for future consistency.

.scalafmt.conf
```diff
@@ -1,45 +1,140 @@
...
+  imports {
+    expand = true
+    sort = scalastyle
+    groups = [
+      ["scala\\.language\\..*",
+        "scala\\.annotation\\..*",
+        "scala\\.reflect\\..*"],
+      ["javax?\\..*",
+        "scala\\..*",
+        "play\\..*"],
```
errt, line 103: Consider adding junit and scalatest before opal

DEVELOPING_OPAL/demos/src/main/scala/org/opalj/br/InvokedynamicPrinter.scala
```diff
@@ -32,12 +31,10 @@ object InvokedynamicPrinter extends ProjectAnalysisApplication {
         libcpFiles:              Iterable[File],
         completelyLoadLibraries: Boolean,
         fallbackConfiguration:   Config
-    )(
-        implicit
-        initialLogContext: LogContext
-    ): Project[URL] = {
+      )(implicit
```
errt, line 34: I guess this is the best we can achieve for implicit parameter lists, which would be setting
`newlines.implicitParamListModifierPrefer = after`
The old behavior would require setting that `Prefer` setting to both, before and after, but it only takes one of them, while there is also a `Force` setting that allows setting both but would always do this, even if it fits into a single line.

Cf. https://www.github.com/scalameta/scalafmt/issues/2279

.scalafmt.conf
errt: https://docs.scala-lang.org/style/declarations.html has a preferred order:
`override`, `private`, `protected`, `implicit`, `final`, `lazy` (the last one is down below, "Lazy vals should use the lazy keyword directly before the val"

This misses `abstract`, but I guess that could be seen as kind of an opposite to `override` and thus should be in the same spot/before that? Alternatively, placing it with `final` could be an option, both can't be set together anyway.

This also misses `sealed`, as that's a modifier only for classes. This, too, could be seen as similar to either `override` or `final`. Closer to `final`, I guess?

I'm not sure I'm happy with placing `implicit` that late, but a short survey showed that others also don't think it's too important, and if there's already an official stance here, maybe we should just go with it.

Any opinions?

DEVELOPING_OPAL/demos/src/main/scala/org/opalj/ai/domain/l1/IfNullParameterAnalysis.scala
roterEmil: What in particular is here not very nice for you?

errt: It forced the map out of braces, making this a single-statement method. Though upon a second look, it's probably fine.

.scalafmt.conf
errt: Okay, since `abstract override` is a thing, `abstract` should be before `override`, and so far we have used `sealed` before `abstract` which seems to be what others do as well, so I'm inclined to put these two in the front. Still not super happy with `implicit` and `final` so late, but a standard is a standard, so let's comply with it.

build.sbt
```diff
@@ -81,9 +77,10 @@ ThisBuild / javaOptions ++= Seq(
 
 addCommandAlias(
   "compileAll",
-  "; copyResources ; scalastyle ; " +
-    "OPAL / Test / compile ; OPAL / Test / scalastyle ; " +
-    "OPAL / IntegrationTest / scalariformFormat ; OPAL / IntegrationTest / scalastyle ; OPAL / IntegrationTest / compile "
+    "OPAL / scalafmt ; OPAL / Test / scalafmt ; " +
```
errt, line 80: I see Test/headerCheck and IntegrationTest/headerCheck, shouldn't there also be OPAL/headerCheck?

build.sbt
```diff
@@ -159,6 +157,7 @@ lazy val `OPAL` = (project in file("."))
 //  .configure(_.copy(id = "OPAL"))
   .settings(Defaults.coreDefaultSettings ++ Seq(publishArtifact := false): _*)
   .enablePlugins(ScalaUnidocPlugin)
+  .disablePlugins(HeaderPlugin)
```
errt, line 160: Why is this?

errt:
Superseded by #182

PR #123, opened 2023-07-13
Clean up CallGraphKey

errt:
Collect the analyses in a single place to remove code duplication and prevent inconsistencies

Free up the initialization data of the key for something more sensible

### DIFF
--- a/OPAL/tac/src/main/scala/org/opalj/tac/cg/CallGraphKey.scala
+++ b/OPAL/tac/src/main/scala/org/opalj/tac/cg/CallGraphKey.scala
@@ -111,7 +111,7 @@ trait CallGraphKey extends ProjectInformationKey[CallGraph, Nothing] {
             implicit val logContext: LogContext = project.logContext
             OPALLogger.error(
                 "analysis configuration",
-                s"call graph was already computed, computing new call graph"
+                s"must not compute multiple call graphs"
             )
             throw new IllegalArgumentException()
         }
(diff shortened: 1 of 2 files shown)


OPAL/tac/src/main/scala/org/opalj/tac/cg/CallGraphKey.scala
```diff
@@ -48,6 +52,10 @@ trait CallGraphKey extends ProjectInformationKey[CallGraph, Nothing] {
     ): Iterable[FPCFAnalysisScheduler]
 
     override def requirements(project: SomeProject): ProjectInformationKeys = {
+        project.updateProjectInformationKeyInitializationData(AIDomainFactoryKey) {
+            _ => Set(classOf[DefaultDomainWithCFGAndDefUse[URL]])
+        }
+
```
Breitfelder, line 58: Why did this code from the test file FieldLocalityTests.scala get into the CallGraphKey.scala?

OPAL/tac/src/main/scala/org/opalj/tac/cg/CallGraphKey.scala
```diff
@@ -102,20 +103,29 @@ trait CallGraphKey extends ProjectInformationKey[CallGraph, Nothing] {
...
+    }
+
+    override def compute(project: SomeProject): CallGraph = {
+        if (CallGraphKey.cg.isDefined && project.availableProjectInformation.contains(CallGraphKey.cg.get)) {
+            implicit val logContext: LogContext = project.logContext
+            OPALLogger.error(
+                "analysis configuration",
+                s"call graphs was already computed, computing new call graph"
+            )
+            throw new IllegalArgumentException()
```
Breitfelder: Typo: call graphs => call graph?

The error message sounds to me like a new call graph is automatically calculated.
Is the program execution interrupted by the following exception, which would prevent this?

OPAL/tac/src/main/scala/org/opalj/tac/cg/CallGraphKey.scala
errt: Because it needs to be done anyway and putting it into the CallGraphKey ensures it is done.

Breitfelder (approved):

PR #116, opened 2023-05-14
Add field access information analysis

maximilianruesch:
Adds an eager analysis for the `FieldAccessInformation` which is currently a `ProjectInformationKey`.

PR #126, opened 2023-07-16
Migrate to new field access information analysis

maximilianruesch:
Migrates (almost) all usages of the old field access information to the new collaborative field access information analysis.

* Depends on #116


DEVELOPING_OPAL/demos/src/main/scala/org/opalj/br/FieldAccessInformationAnalysis.scala
```diff
@@ -54,14 +55,21 @@ object FieldAccessInformationAnalysis extends ProjectAnalysisApplication {
             val Array(declaringClassName, fieldName) =
                 parameters.head.substring(7).replace('.', '/').split(' ')
             val declaringClassType = ObjectType(declaringClassName)
-            val writes = accessInformation.writeAccesses(declaringClassType, fieldName)
-            val reads = accessInformation.readAccesses(declaringClassType, fieldName)
+            val fields = project.classFile(declaringClassType).map(_.findField(fieldName)).getOrElse(List.empty)
+
+            val (reads, writes) = fields
+                .foldLeft((Seq.empty[(DefinedMethod, PCs)], Seq.empty[(DefinedMethod, PCs)])) { (accesses, field) =>
+                    val newReads = (accesses._1 ++ accessInformation.readAccesses(field)).asInstanceOf[Seq[(DefinedMethod, PCs)]]
```
errt, line 62: These type casts can't work, readAccesses returns a four-tuple

DEVELOPING_OPAL/demos/src/main/scala/org/opalj/fpcf/analyses/FieldAccessInformationAnalysisDemo.scala
```diff
@@ -0,0 +1,115 @@
...
+        }
+
+        var propertyStore: PropertyStore = null
+        var analysisTime: Seconds = Seconds.None
+        val analysesManager = project.get(FPCFAnalysesManagerKey)
+
+        time {
+            propertyStore = analysesManager
+                .runAll(
+                    AllocationSiteBasedPointsToAnalysisScheduler,
```
errt: How can this work without a call graph?

OPAL/ai/src/main/scala/org/opalj/ai/fpcf/analyses/LBFieldValuesAnalysis.scala
```diff
@@ -430,7 +439,12 @@ object FieldValuesAnalysis {
 
 object EagerLBFieldValuesAnalysis extends BasicFPCFEagerAnalysisScheduler {
 
-    override def requiredProjectInformation: ProjectInformationKeys = Seq(FieldAccessInformationKey)
+    override def requiredProjectInformation: ProjectInformationKeys = Seq(
+        FieldAccessInformationKey,
+        DeclaredMethodsKey,
+        DeclaredFieldsKey,
+        ContextProviderKey // TODO type mismatch (only in IDE) here?
```
errt, line 446: The reasons seems to be that both DeclaredMethodsKey and DeclaredFieldsKey have `Nothing` for the second generic type. With only one key having that, the type is still correctly inferred as `:< Object`, but with both, the bound is not correctly inferred anymore.
I don't fully understand this. `Nothing` is at the bottom of Scala's type hierarchy, so it is a subtype of AnyRef (Any even). The same problem occurs if both keys have `Null` (also at the bottom of the type hierarchy, but only subtype of AnyRef) as the generic type or one key has `Nothing` or `Null`, but not if only one key has `Nothing` or `Null` and the others have actual AnyRef types.

OPAL/br/src/main/scala/org/opalj/br/analyses/DeclaredFields.scala
```diff
@@ -0,0 +1,104 @@
...
+package analyses
+
+import org.opalj.br.instructions.FieldAccess
+import org.opalj.log.OPALLogger.info
+
+import java.util.concurrent.ConcurrentHashMap
+import java.util.concurrent.atomic.AtomicInteger
+import java.util.concurrent.locks.ReentrantReadWriteLock
+
+class DeclaredFields(
```
errt: Document & add author information

OPAL/br/src/main/scala/org/opalj/br/analyses/DeclaredFieldsKey.scala
```diff
@@ -0,0 +1,37 @@
+/* BSD 2-Clause License - see OPAL/LICENSE for details. */
+package org.opalj
+package br
+package analyses
+
+import java.util.concurrent.ConcurrentHashMap
+import java.util.concurrent.atomic.AtomicInteger
+import scala.jdk.CollectionConverters.EnumerationHasAsScala
+
+object DeclaredFieldsKey extends ProjectInformationKey[DeclaredFields, Nothing] {
```
errt: Document & add author information

OPAL/br/src/main/scala/org/opalj/br/analyses/FieldAccessInformation.scala
```diff
@@ -3,82 +3,70 @@ package org.opalj
...
 
-    final def readAccesses(field: Field): Seq[(Method, PCs)] = {
-        allReadAccesses.getOrElse(field, Seq.empty)
-    }
-
-    def isRead(field: Field): Boolean = {
-        allReadAccesses.get(field) match {
-            case Some(accesses) => accesses.nonEmpty
-            case None           => false
+    private[this] def getFieldAccessInformation[S <: fieldaccess.FieldAccessInformation[S]](
```
errt, line 28: Should the result of this be cached?

errt (changes requested):
Review, first half

OPAL/tac/src/main/scala/org/opalj/tac/fpcf/analyses/cg/xta/TypePropagationAnalysis.scala
```diff
@@ -265,6 +275,72 @@ final class TypePropagationAnalysis private[analyses] (
...
+                // (but not all!), the Java compiler will place the "actual" return type within a checkcast
+                // instruction right after the field read instruction.
+                val nextInstruction = bytecode.instructions(bytecode.pcOfNextInstruction(pc))
+                val mostPreciseFieldType =
+                    if (nextInstruction.isCheckcast)
+                        nextInstruction.asInstanceOf[CHECKCAST].referenceType
+                    else
+                        declaredField.fieldType.asReferenceType
+
+                declaredField match {
```
errt, line 304: Is this match still required? It both cases, it just passes `(declaredField, mostPreciseFieldType)` to `registerEntityForBackwardPropagation` (`asDefinedField` is a no-op on DefinedFields and `registerEntityForBackwardPropagation` is not overloaded for DefinedField vs. DeclaredField).

OPAL/tac/src/main/scala/org/opalj/tac/fpcf/analyses/cg/xta/TypePropagationAnalysis.scala
```diff
@@ -265,6 +275,72 @@ final class TypePropagationAnalysis private[analyses] (
...
+            declaredField <- writeAccesses.getNewestAccessedFields(
+                state.callContext,
+                pc,
+                state.seenDirectWriteAccesses.getOrElse(pc, 0),
+                state.seenIndirectWriteAccesses.getOrElse(pc, 0)
+            )
+        } {
+            val fieldType = declaredField.fieldType
+            if (fieldType.isReferenceType) {
+                declaredField match {
```
errt, line 334: As above, the match seems unnecessary

OPAL/tac/src/main/scala/org/opalj/tac/fpcf/analyses/cg/xta/TypePropagationAnalysis.scala
```diff
@@ -265,6 +275,72 @@ final class TypePropagationAnalysis private[analyses] (
...
+            pc <- readAccesses.getAccessSites(state.callContext)
+            declaredField <- readAccesses.getNewestAccessedFields(
+                state.callContext,
+                pc,
+                state.seenDirectReadAccesses.getOrElse(pc, 0),
+                state.seenIndirectReadAccesses.getOrElse(pc, 0)
+            )
+        } {
+            if (declaredField.fieldType.isReferenceType) {
+                // Internally, generic fields have type "Object" due to type erasure. In many cases
```
errt, line 294: I wonder how this applies to indirect field accesses, but it is probably still safe to use a more precise type from a checkcast anyway.

OPAL/tac/src/main/scala/org/opalj/tac/fpcf/analyses/cg/xta/TypeSetEntitySelector.scala
```diff
@@ -30,8 +31,8 @@ object XTASetEntitySelector extends TypeSetEntitySelector {
     override def apply(e: Entity): TypeSetEntity = e match {
         case dm: DefinedMethod        => dm
         case _: VirtualDeclaredMethod => ExternalWorld
-        case f: Field                 => f
-        case _: ExternalField         => ExternalWorld
+        case f: DefinedField          => f.definedField
```
errt, line 34: As with DefinedMethod, we could probably attach the type sets to the DefinedField directly instead of the Field. That would probably cause further changes in the analyses, though.

TOOLS/hermes/src/main/scala/org/opalj/hermes/queries/MicroPatterns.scala
```diff
@@ -258,13 +264,16 @@ class MicroPatterns(implicit hermes: HermesConfig) extends FeatureQuery {
...
+        contextProvider: ContextProvider,
+        declaredFields:  DeclaredFields
+    ): Boolean = {
         !cl.isInterfaceDeclaration &&
             !cl.isAbstract && cl.fields.count { f => !f.isStatic } > 1 &&
             cl.fields.forall(f => f.isPrivate && !f.isStatic) &&
             cl.fields.forall { f =>
-                !fa.allWriteAccesses.contains(f) ||
-                    fa.allWriteAccesses(f).forall(p => isInitMethod(p._1))
+                fa.writeAccesses(f).forall(p => isInitMethod(contextProvider.contextFromId(p._1).method.definedMethod))
```
errt, line 276: Could this lead to an exception or are we always guaranteed to have (write)accesses available for every field?

OPAL/tac/src/main/scala/org/opalj/tac/fpcf/analyses/FieldLocalityState.scala
```diff
@@ -47,14 +53,18 @@ class FieldLocalityState(val field: Field, val thisIsCloneable: Boolean) {
...
+            fieldWriteAccessDependee.filter(_.isRefinable) ++
             definitionSitesDependees.valuesIterator.map(_._1) ++
             tacDependees.valuesIterator.filter(_.isRefinable) ++
             callerDependees.valuesIterator.filter(_.isRefinable) ++
             calleeDependees.valuesIterator.map(_._1).filter(_.isRefinable)).toSet
     }
 
     def hasNoDependees: Boolean = {
-        tacDependees.valuesIterator.forall(_.isFinal) &&
+        fieldReadAccessDependee.forall(_.isRefinable) &&
```
errt: Shouldn't these be `_.isFinal`? The method is hasNOdependees.

errt (changes requested):
Review, part 2

OPAL/tac/src/main/scala/org/opalj/tac/fpcf/analyses/FieldLocalityAnalysis.scala
```diff
@@ -582,25 +570,198 @@ class FieldLocalityAnalysis private[analyses] (
...
+
+            faiEP.ub.getNewestAccesses(
+                faiEP.ub.numDirectAccesses - seenDirectAccesses,
+                faiEP.ub.numIndirectAccesses - seenIndirectAccesses
+            ) exists { wa =>
+                    val definedMethod = contextProvider.contextFromId(wa._1).method
+                    val method = definedMethod.definedMethod
+                    val pc = wa._2
+                    val parameter = wa._4
+                    state.tacFieldWriteAccesses += method -> (state.tacFieldWriteAccesses.getOrElse(method, Set.empty) + ((pc, parameter)))
```
errt: Line length

OPAL/tac/src/main/scala/org/opalj/tac/fpcf/analyses/FieldLocalityAnalysis.scala
```diff
@@ -698,25 +825,23 @@ class FieldLocalityAnalysis private[analyses] (
...
-            PropertyBounds.ub(ReturnValueFreshness),
-            PropertyBounds.ub(Callees)
-        )
-    }
+    final override def uses: Set[PropertyBounds] = PropertyBounds.ubs(
+        TACAI,
+        EscapeProperty,
+        ReturnValueFreshness,
+        Callees,
+        FieldWriteAccessInformation
```
errt, line 843: FieldReadAccessInformation is also used

OPAL/tac/src/main/scala/org/opalj/tac/fpcf/analyses/fieldassignability/L0FieldAssignabilityAnalysis.scala
```diff
@@ -70,44 +93,82 @@ class L0FieldAssignabilityAnalysis private[analyses] (val project: SomeProject)
...
+            ) exists { wa =>
+                    val method = contextProvider.contextFromId(wa._1).method.definedMethod
+                    if (method.isStaticInitializer) {
+                        if (wa._3.isDefined) {
+                            // If a receiver is defined, we know that the access was not static
+                            // IMPROVE: Add static information to accesses and resolve this
+                            false
+                        } else {
+                            // As a fallback, we look for a trivial static write access in the statements
+                            method.body.get.instructions(wa._2).opcode == PUTSTATIC.opcode
```
errt: Is this sound? If we lack information, we should probably always fall back to assuming that the field could be assignable.

OPAL/tac/src/main/scala/org/opalj/tac/fpcf/analyses/fieldassignability/L2FieldAssignabilityAnalysis.scala
```diff
@@ -895,32 +995,17 @@ class L2FieldAssignabilityAnalysis private[analyses] (val project: SomeProject)
...
-        DeclaredMethodsKey,
-        FieldAccessInformationKey,
-        ClosedPackagesKey,
-        TypeExtensibilityKey,
-        DefinitionSitesKey,
-        ContextProviderKey
+trait L2FieldAssignabilityAnalysisScheduler extends AbstractFieldAssignabilityAnalysisScheduler {
+    override def uses: Set[PropertyBounds] = super.uses ++ PropertyBounds.ubs(
+        FieldReadAccessInformation,
+        FieldWriteAccessInformation
```
errt: FieldWriteAccessInformation is already specified in the abstract scheduler, no need to repeat it here.

OPAL/tac/src/main/scala/org/opalj/tac/fpcf/analyses/cg/TypeIterator.scala
```diff
@@ -907,45 +925,54 @@ abstract class AbstractAllocationSitesPointsToTypeIterator(project: SomeProject)
...
+                    )
+                    theTAC <- tacEP.tac
+                    defSite <- uVarForDefSites(receiver.get, theTAC.pcToIndex).definedBy
+                } {
+                    val defPC = if (defSite < 0) defSite else theTAC.stmts(defSite).pc
+
+                    result = combine(
+                        result,
+                        typesProperty(
+                            field, DefinitionSite(method, defPC), depender, newContext(definedMethod), theTAC.stmts
```
errt: Line length

OPAL/tac/src/main/scala/org/opalj/tac/fpcf/analyses/cg/TypeIterator.scala
```diff
@@ -1082,40 +1146,55 @@ class CFA_k_l_TypeIterator(project: SomeProject, val k: Int, val l: Int)
...
+
+                result
             }
+        } else {
+            NoAllocationSites // FIXME: Deal with VirtualDeclaredFields
         }
     }
 
-    // TODO several field-related methods are missing here!
+    // TODO several field-related methods are missing here! Implement this when field access information considers reflective calls
```
errt: Now that we have reflective calls covered in #145, what about this?

OPAL/tac/src/main/scala/org/opalj/tac/fpcf/analyses/cg/TypeIterator.scala
```diff
@@ -1082,40 +1146,55 @@ class CFA_k_l_TypeIterator(project: SomeProject, val k: Int, val l: Int)
...
+                    result = combine(
+                        result,
+                        typesProperty(field, DefinitionSite(method, defPC), depender, calleeContext, theTAC.stmts)
+                    )
                 }
+
+                result
             }
+        } else {
+            NoAllocationSites // FIXME: Deal with VirtualDeclaredFields
```
errt: What about this?

OPAL/tac/src/main/scala/org/opalj/tac/fpcf/analyses/pointsto/AbstractPointsToAnalysis.scala
```diff
@@ -202,6 +200,7 @@ trait AbstractPointsToAnalysis extends PointsToAnalysisBase with ReachableMethod
                     )
                 }
                 handleCall(call, pc)
+                handleIndirectFieldAccesses(pc)
```
errt: As this requires no information about the call, why is it invoked for each call individually instead of just once globally for all indirect accesses in the method?

OPAL/tac/src/main/scala/org/opalj/tac/fpcf/analyses/pointsto/AbstractPointsToAnalysis.scala
```diff
@@ -543,11 +632,50 @@ trait AbstractPointsToAnalysis extends PointsToAnalysisBase with ReachableMethod
 
                 Results(results)
 
+            case UBP(fai: MethodFieldReadAccessInformation) =>
+                val defSite = e match {
+                    case ds: DefinitionSite               => ds
+                    case (_: Context, ds: DefinitionSite) => ds
+                }
+                // IMPROVE: Only handle new accesses
```
errt, line 646: Can we do this?

OPAL/tac/src/main/scala/org/opalj/tac/fpcf/analyses/pointsto/AbstractPointsToAnalysis.scala
```diff
@@ -588,6 +716,38 @@ trait AbstractPointsToAnalysis extends PointsToAnalysisBase with ReachableMethod
...
+                val newAccesses = if (eps.hasUBP) eps.ub.asInstanceOf[P] else noAccesses
+                val oldAccesses = if (oldAccessEOptP.hasUBP) oldAccessEOptP.ub else noAccesses
+
+                val tac = state.tac
+                for {
+                    pc <- newAccesses.getIndirectAccessSites(state.callContext)
+                    target <- newAccesses.getNewestNIndirectAccessedFields(
+                        state.callContext,
+                        pc,
+                        newAccesses.numIndirectAccesses(state.callContext, pc) - oldAccesses.numIndirectAccesses(state.callContext, pc)
```
errt: Line length

OPAL/tac/src/main/scala/org/opalj/tac/fpcf/analyses/pointsto/AbstractPointsToAnalysis.scala
```diff
@@ -597,7 +757,7 @@ trait AbstractPointsToAnalysisScheduler extends FPCFTriggeredAnalysisScheduler {
     override type InitializationData = Null
 
     override def requiredProjectInformation: ProjectInformationKeys =
-        Seq(DeclaredMethodsKey, VirtualFormalParametersKey, DefinitionSitesKey, TypeIteratorKey)
+        Seq(DeclaredMethodsKey, DeclaredFieldsKey, VirtualFormalParametersKey, DefinitionSitesKey, TypeIteratorKey)
 
     override def uses: Set[PropertyBounds] = PropertyBounds.ubs(
         Callers,
```
errt, line 770: Field accesses?

errt (changes requested):
Review, last part

OPAL/tac/src/main/scala/org/opalj/tac/fpcf/analyses/fieldassignability/L0FieldAssignabilityAnalysis.scala
maximilianruesch: In theory, we could look if we encounter a trivial `PUTSTATIC`, otherwise assume its assignable...

But that is just the same as returning true here anyways. Will do.

OPAL/tac/src/main/scala/org/opalj/tac/fpcf/analyses/cg/TypeIterator.scala
maximilianruesch: I guess (guessing as i did not place this FIXME myself) we could return `currentPointsTo` in case the field is defined and static. Information about fields being static is not yet persisted for VirtualDeclaredFields, else we could generalize this.
The else part (currently: the field is defined but an instance field) we could probably just also allow for virtual declared fields.

OPAL/tac/src/main/scala/org/opalj/tac/fpcf/analyses/cg/TypeIterator.scala
maximilianruesch: In theory this is still open. We will need to shortly discuss what we need here and add it in another PR (as to not make the existing ones any bigger than they already are). In the meantime, i will revert the change here, as it wont make sense after #145.

OPAL/tac/src/main/scala/org/opalj/tac/fpcf/analyses/pointsto/AbstractPointsToAnalysis.scala
maximilianruesch: How do i know which indirect accesses there are in the method? I.e. which Type do they have in common?

OPAL/tac/src/main/scala/org/opalj/tac/fpcf/analyses/pointsto/AbstractPointsToAnalysis.scala
maximilianruesch: Theoretically, yes. However as this is in the `continuationForShared`, i am quite unsure how to best track "old" eoptp or rather seen accesses per PC independently of the other continuation.

OPAL/tac/src/main/scala/org/opalj/tac/fpcf/analyses/cg/TypeIterator.scala
maximilianruesch: According to my IDE, this is still a little under 120 chars. I will break it up, but for future cases: Should there be a line length limit other than 120 characters be configured?

OPAL/tac/src/main/scala/org/opalj/tac/fpcf/analyses/cg/TypeIterator.scala
```diff
@@ -959,7 +959,8 @@ class AllocationSitesPointsToTypeIterator(project: SomeProject)
                     result = combine(
                         result,
                         typesProperty(
-                            field, DefinitionSite(method, defPC), depender, newContext(definedMethod), theTAC.stmts
+                            field, DefinitionSite(method, defPC),
```
errt: One parameter per line please

OPAL/tac/src/main/scala/org/opalj/tac/fpcf/analyses/cg/TypeIterator.scala
```diff
@@ -934,41 +934,38 @@ class AllocationSitesPointsToTypeIterator(project: SomeProject)
...
-                    result = combine(
-                        result,
-                        typesProperty(
-                            field, DefinitionSite(method, defPC),
-                            depender, newContext(definedMethod), theTAC.stmts
-                        )
-                    )
-                }
+        if (field.isDefinedField && field.definedField.isStatic) {
+            // IMPROVE: Handle static case also for VirtualDeclaredFields when static information is available on them
```
errt: Should this be "once static information is available"?

OPAL/tac/src/main/scala/org/opalj/tac/fpcf/analyses/pointsto/AbstractPointsToAnalysis.scala
errt: `state.readAccesses.getIndirectAccessSites`?

OPAL/tac/src/main/scala/org/opalj/tac/fpcf/analyses/cg/TypeIterator.scala
errt: If it is, keep it. Looked longer to me, but I didn't check exactly.

OPAL/tac/src/main/scala/org/opalj/tac/fpcf/analyses/pointsto/AbstractPointsToAnalysis.scala
errt: `dependees(eps.toEPK)._1` should give you the previous state of the field accesses. Not sure about about per-PC though, that probably doesn't work right now. If it required large changes, let's keep it like it is.

maximilianruesch:
The tests are running now... What a marvel.

OPAL/tac/src/main/scala/org/opalj/tac/fpcf/analyses/cg/TypeIterator.scala
maximilianruesch: Obsolete as per https://github.com/opalj/opal/pull/126#discussion_r1375876127

OPAL/tac/src/main/scala/org/opalj/tac/fpcf/analyses/pointsto/AbstractPointsToAnalysis.scala
maximilianruesch: holy**** (i literally wrote the code myself)

OPAL/tac/src/main/scala/org/opalj/tac/fpcf/analyses/pointsto/AbstractPointsToAnalysis.scala
maximilianruesch: It requires some changes in the state to track already seen accesses on a per-PC basis. It should not be too hard to implement, just to get it right and understandable. Lets shortly discuss this next meeting.

OPAL/tac/src/main/scala/org/opalj/tac/fpcf/analyses/pointsto/AbstractPointsToAnalysis.scala
```diff
@@ -462,16 +462,22 @@ trait AbstractPointsToAnalysis extends PointsToAnalysisBase with ReachableMethod
...
+        for (
+            pc <- readAccesses.getAccessSites(state.callContext);
+            target <- readAccesses.indirectAccessedFields(state.callContext, pc)
+        ) {
             handleIndirectFieldReadAccess(pc, target, readAccesses, tac)
         }
 
-        for (target <- writeAccesses.indirectAccessedFields(state.callContext, pc)) {
+        for (
+            pc <- readAccesses.getAccessSites(state.callContext);
```
errt: Should most likely be `writeAccesses`

errt:
The reason for the failing tests is that LongMap, used to store field access information, is not ordered and, thus, `.iterator.take(n)` in `getNewestNDirectAccesses` is not guaranteed to return the newest accesses.

maximilianruesch:
How did you get to this result?

errt:
Put some `println` in strategic places, managed to reproduce the error by executing the tests using sbt directly instead of running them in IntelliJ. This showed that when the tests fail, the two write accesses to the field are both recognized, the continuation is triggered and the newest accesses are accessed twice, each time requesting one access, but getting back the same one each time. From there, I looked up how the newest accesses are computed and when I saw the maps, it was pretty obvious.

maximilianruesch:
In the in-person reviews we thought of the LongMaps as being ordered by entry modification timestamp. I remember that at some point we were not quire sure though as the implementation is quite messy as far as i remember.

Should we revert the longmaps to the double IntMap like the Implementation from the reverse side?

errt:
Can you fix this?

maximilianruesch:
> Can you fix this?

If you answer the following question as positive, yes, otherwise, unsure
> Should we revert the `LongMap` to the double`IntMap` like the Implementation from the reverse side?

errt:
Pretty sure `IntMap` isn't ordered as well. `Callees` gets away using `IntMap` because it has no methods to access the newest entries, `AllocationSitePointsToSet` uses `LongLinkedSet` which is ordered by virtue of the linking.

maximilianruesch:
We should then fix this for both types of `FieldAccessInformation`. I will have a look into this!

Thanks for your work in debugging this.

maximilianruesch:
I am right now pondering about the outward facing interface of the FieldAccessInformation Classes. They are kind of forced now to take some kind of ordered set for constructing of new instances. In the case of the `MethodFieldAccessInformation` this should not be a problem, as the accesses are not encoded (still in `Int` form).

For the `FieldAccessInformation` however, accesses are encoded as `Long` which simplifies the iteration API (the getNewestX functions). Requiring an ordered long set for construction however also requires exposing the `encodeFieldAccesses` as it will have to be used in all places requiring construction of such objects. We can say that is fine, as we construct them all through `FieldAccesses` anyways and can close off access to this function to inside the package.

The alternative is rebuilding the iteration API to match the one of the `MethodFieldAccessInformation` and take IntMaps (or their ordered equivalent) instead.

What is your opinion on this? Imo we can do the first one and just close off the encoding functions.

errt:
The first approach sounds reasonable, let's try that.

maximilianruesch:
It seems this at least does not fail the CI anymore (the one before was flaky, but yknow what i mean)

errt:
I don't think the changes in `MethodFieldAccessInformation` are sufficient. `getNewestAccessedFields` still uses the contents of `_(in)directAccessedReceiversByField` and that isn't ordered.

maximilianruesch:
:bruh: i was so entangled in updating the structure, i forgot to actually use it :sweat_smile:

maximilianruesch:
The reason this fails is that previously we were effectively using a `Set` (by virtue of just having Intmap keys) that was unordered and now we are using a `List` that is ordered by insertion, but lose the uniqueness property of `Set`.

I think the best way forward (in the long run) would be to reimplement `LongLinkedSet` for integers. Do you have any ideas of collections that are usable here to avoid this task for this PR?

errt:
I don't have any ideas for other collections. I guess it's using LongLinkedTrieSet for now, or copy that and adapt it to Int.

maximilianruesch:
I found the reason for the test failure: In `FieldAccesses`, the accessed fields which are as of the newest commits actually used in the MethodFieldAccessInformation were not recorded individually for reads and writes. This caused some weird cross-contamination type stuff. Fixed it by seperating these.

I also reverted the switch to `LongLinkedSet` and it seems to work fine (locally and on the CI).

OPAL/br/src/main/scala/org/opalj/br/fpcf/properties/fieldaccess/FieldAccesses.scala
```diff
@@ -174,15 +179,15 @@ trait CompleteFieldAccesses extends FieldAccesses {
...
+            _accessedFields = _accessedFields.updated(pc, IntList(field.id))
             _partialResultsForFieldBasedFieldAccesses ::= createFieldPartialResultForContext(
                 field,
                 propertyKey,
                 propertyFactory()
             )
         } else {
             val oldFieldsAtPC = oldFieldsAtPCOpt.get
-            val newFieldsAtPC = oldFieldsAtPC + field.id
+            val newFieldsAtPC = oldFieldsAtPC.+:(field.id)
```
errt: Use as infix operator

OPAL/br/src/main/scala/org/opalj/br/fpcf/properties/fieldaccess/MethodFieldAccessInformation.scala
```diff
@@ -177,6 +180,22 @@ sealed trait MethodFieldAccessInformation[S <: MethodFieldAccessInformation[S]]
...
+        contextId:   Int,
+        updateValue: IntMap[IntList]
+    ): IntMap[IntMap[IntList]] = {
+        baseMap.updateWith(
+            contextId,
+            updateValue,
+            (o, n) =>
+                o.unionWith(
+                    n,
+                    (_, l, r) => l.++:(r)
```
errt: Use as infix operator

OPAL/br/src/main/scala/org/opalj/br/fpcf/properties/fieldaccess/FieldAccesses.scala
```diff
@@ -224,15 +214,15 @@ trait CompleteFieldAccesses extends FieldAccesses {
...
+            _writeFields = _writeFields.updated(pc, IntList(field.id))
             _partialResultsForFieldBasedFieldAccesses ::= createFieldPartialResultForContext(
                 field,
                 FieldWriteAccessInformation.key,
                 propertyFactory()
             )
         } else {
             val oldFieldsAtPC = oldFieldsAtPCOpt.get
-            val newFieldsAtPC = oldFieldsAtPC + field.id
+            val newFieldsAtPC = oldFieldsAtPC.+:(field.id)
```
errt: Use as infix operator

OPAL/common/src/main/scala/org/opalj/collection/immutable/IntList.scala
```diff
@@ -129,6 +133,15 @@ final case class IntListNode(
     }
 
     override def +:(v: Int): IntList = new IntListNode(v, this)
+    override def ++:(other: IntList): IntList = {
+        var newList = this
```
errt: Replace with a fold?

errt (changes requested):

OPAL/tac/src/main/scala/org/opalj/tac/fpcf/analyses/pointsto/AbstractPointsToAnalysis.scala
errt: Did we reach a conclusion on this?

errt (approved):

PR #131, opened 2023-07-27
Create run_it_test_on_comment.yml

TorunR:
Create a new workflow that is triggered when a PR is commented with "/it:test".  This will only run on self-hosted runners with the "it_test" label.

### DIFF
--- a/.github/workflows/run_it_test_on_comment.yml
+++ b/.github/workflows/run_it_test_on_comment.yml
@@ -1,0 +1,33 @@
+name: "Run it:test on Comment"
+
+on:
+  issue_comment:
+    types: [created]
+
+jobs:
+  build:
+    runs-on: [self-hosted, it_test]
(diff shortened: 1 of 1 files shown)


.github/workflows/run_it_test_on_comment.yml
```diff
@@ -0,0 +1,33 @@
...
+jobs:
+  build:
+    runs-on: [self-hosted, it_test]
+
+    if: github.event.issue.pull_request && contains(github.event.comment.body, '/it:test')
+
+    steps:
+
+      - name: Get PR branch
+        uses: xt0rted/pull-request-comment-branch@v1
```
errt: Is there any security implication of running someone else's actions on our machines? Also, why does this use v1 when there seems to be v2?

errt (approved):

PR #142, opened 2023-09-22
Split context functions from TypeIterator

errt:
The `ContextProvider` interface now allows using context functions in `br` and provides better separation of concerns as many clients don't need the full `TypeIterator`.


maximilianruesch:
This PR has some issues related to computing the context provider key when no type iterator is configured. I have added some logging to the initialization of the project variable in the fallback instance and some to the declared methods and simple context variables:
image

1. The project variable is computed after the declared methods and simple context variables (upper part of the screenshot), which leads to a null pointer upon `project.get` in the latter variable initialization
2. Circumventing this temporarily by early returning null will uncover another problem: The context provider is initialized earlier than the type iterator key and thus the fallback CHA type iterator is not used for the context provider which in turn fails the class cast at the bottom of the type iterator

This will need some more work with regards to the fallback. This was tested using the field assignability analysis and commenting out the `RTACallGraphKey`.

*I also suggest to add an `apply` interface if possible*

maximilianruesch:
I can confirm the recent changes fixed the problems with context provider initialization. Analyses now work even without explicitly specifying a context provider, using the SimpleContextProvider as a fallback.

roterEmil (approved):

PR #145, opened 2023-10-04
Add reflective field access analysis

maximilianruesch:
Adds the `EagerReflectionRelatedFieldAccessesAnalysisScheduler` which can extract field accesses via reflective methods auch as `findStaticGetter().invokeExact()`.

Note that this PR depends on both:
1. #116 
2. #126


OPAL/tac/src/main/scala/org/opalj/tac/fpcf/analyses/fieldaccess/reflection/FieldMatcher.scala
```diff
@@ -0,0 +1,150 @@
...
+
+    def initialFields(implicit p: SomeProject): Iterator[Field]
+    def contains(field: Field)(implicit p: SomeProject): Boolean
+    def priority: Int
+}
+
+final class NameBasedFieldMatcher(val possibleNames: Set[String]) extends FieldMatcher {
+
+    override def initialFields(implicit p: SomeProject): Iterator[Field] = {
+        possibleNames.iterator.flatMap(p.get(ProjectIndexKey).findFields)
```
errt: Please only get the key once

OPAL/tac/src/main/scala/org/opalj/tac/fpcf/analyses/fieldaccess/reflection/FieldMatcher.scala
```diff
@@ -0,0 +1,150 @@
...
+
+class ClassBasedFieldMatcher(
+        val possibleClasses:          Set[ObjectType],
+        val onlyFieldsExactlyInClass: Boolean
+) extends FieldMatcher {
+
+    // TODO use a ProjectInformationKey or WeakHashMap to cache fields per project
+    // (for the contains check)
+    private[this] def fields(implicit p: SomeProject): Set[Field] = possibleClasses.flatMap { c =>
+        // todo what about "inherited" fields?
```
errt: Yes, what about them?

OPAL/tac/src/main/scala/org/opalj/tac/fpcf/analyses/fieldaccess/reflection/FieldMatcher.scala
```diff
@@ -0,0 +1,150 @@
...
+        p.classFile(c).map(_.fields).getOrElse(ArraySeq.empty)
+    }
+
+    override def initialFields(implicit p: SomeProject): Iterator[Field] = fields.iterator
+    override def contains(f: Field)(implicit p: SomeProject): Boolean = fields.contains(f)
```
errt, line 52: I wonder if it would be more efficient to use the field's declaring class file here and check whether the possible classes contain it (or, if we consider inherited fields, any of them is a subclass)

OPAL/tac/src/main/scala/org/opalj/tac/fpcf/analyses/fieldaccess/reflection/FieldMatcher.scala
```diff
@@ -0,0 +1,150 @@
...
+ * bound type. Usually encountered in parameters of field accesses.
+ */
+class LBTypeBasedFieldMatcher(val fieldType: FieldType) extends FieldMatcher {
+
+    override def initialFields(implicit p: SomeProject): Iterator[Field] = p.allFields.iterator.filter(contains)
+    override def contains(f: Field)(implicit p: SomeProject): Boolean = isTypeCompatible(f.fieldType, fieldType)(p.classHierarchy)
+    override def priority: Int = 3
+}
+
+class ActualReceiverBasedFieldMatcher(val receiver: IsReferenceValue) extends FieldMatcher {
```
errt, line 84: Document

OPAL/tac/src/main/scala/org/opalj/tac/fpcf/analyses/fieldaccess/reflection/FieldMatcher.scala
```diff
@@ -0,0 +1,150 @@
...
+    override def contains(f: Field)(implicit p: SomeProject): Boolean = {
+        val isNull = receiver.isNull
+        (isNull.isNoOrUnknown && receiver.isValueASubtypeOf(f.classFile.thisType)(p.classHierarchy).isYesOrUnknown) ||
+            (isNull.isYesOrUnknown && f.isStatic)
+    }
+
+    override def priority: Int = 3
+}
+
+class ActualParameterBasedFieldMatcher(val actualParam: V) extends FieldMatcher {
```
errt, line 101: Document

OPAL/tac/src/main/scala/org/opalj/tac/fpcf/analyses/fieldaccess/reflection/ReflectionRelatedFieldAccessesAnalysis.scala
```diff
@@ -0,0 +1,1055 @@
...
+                            tpe => if (tpe.isObjectType) tpe.asObjectType else ObjectType.Object
+                        }),
+                        data.pc,
+                        v => new ClassBasedFieldMatcher(v, !data.matchers.contains(PublicFieldMatcher))
+                    )
+
+                addFieldAccess(state.callContext, data.pc, _ => data.receiver, _ => data.param, matchers)
+            }
+
+        AllocationsUtil.continuationForAllocation[(ClassDepender, V, Array[Stmt[V]]), ContextType](
```
errt: What is this continuation for? It looks broken, since the given type is a 3-tuple, but the instanceOf check checks for a 2-tuple.

OPAL/tac/src/main/scala/org/opalj/tac/fpcf/analyses/fieldaccess/reflection/ReflectionRelatedFieldAccessesAnalysis.scala
```diff
@@ -0,0 +1,1055 @@
...
+                        NameDepender(accessPC, actualReceiver, actualParameter, matchers, params.head.asVar, stmts, receiver.asVar, context),
+                        accessPC,
+                        stmts,
+                        () => failure(accessPC, actualReceiver, actualParameter, matchers)
+                    )
+
+                if (!matchers.contains(NoFieldsMatcher))
+                    matchers += constructClassBasedFieldMatcherForClassVar(receiver.asVar, matchers, !isGetField)
+
+            case call @ VirtualFunctionCall(_, ObjectType.Class, _, "getDeclaredFields" | "getFields", _, receiver, _) =>
```
errt: I'm still unsure whether it is sensible to handle this here. This can never happen on the initial call of `handleGetField`, only for the recursion in the ArrayLoad case, right?

OPAL/tac/src/main/scala/org/opalj/tac/fpcf/analyses/fieldaccess/reflection/ReflectionRelatedFieldAccessesAnalysis.scala
```diff
@@ -0,0 +1,1055 @@
...
+        AllocationsUtil.handleAllocations(
+            fieldVar, accessContext, depender, state.tac.stmts, _ eq ObjectType.Field,
+            () => failure(accessPC, persistentReceiver, None, baseMatchers)
+        ) { (allocationContext, allocationIndex, stmts) =>
+                val allMatchers = handleGetField(
+                    allocationContext, accessPC, allocationIndex,
+                    persistentReceiver, None,
+                    baseMatchers, stmts
+                )
+                addFieldRead(
```
errt: Probably fits in a single line

OPAL/tac/src/main/scala/org/opalj/tac/fpcf/analyses/fieldaccess/reflection/ReflectionRelatedFieldAccessesAnalysis.scala
```diff
@@ -0,0 +1,1055 @@
...
+                    retrieveSuitableMatcher[Set[ObjectType]](
+                        Some(classes),
+                        data.pc,
+                        v => new ClassBasedFieldMatcher(v, false)
+                    )
+
+                addFieldAccesses(state.callContext, data.pc, matchers, data.persistentActualParams)
+            }
+
+        AllocationsUtil.continuationForAllocation[(ClassDepender, V, Array[Stmt[V]]), ContextType](
```
errt: What is this continuation for? Why does it use a tuple type instead of a case class like the others?

OPAL/tac/src/main/scala/org/opalj/tac/fpcf/analyses/fieldaccess/reflection/ReflectionRelatedFieldAccessesAnalysis.scala
```diff
@@ -0,0 +1,1055 @@
...
+        }
+
+        matchers
+    }
+}
+
+class FieldGetAnalysis private[analyses] (
+        final val project:       SomeProject,
+        final val apiMethodName: String,
+        final val fieldType:     Option[BaseType] = None
```
errt: Since this is only ever used with the default argument value once, maybe remove the default and supply it explicitly in that single case as well for better clarity.

errt (changes requested):
Partial review

OPAL/tac/src/main/scala/org/opalj/tac/fpcf/analyses/fieldaccess/reflection/ReflectionRelatedFieldAccessesAnalysis.scala
maximilianruesch: Well if also foreach loops over arrays are compiled to array loads (which would make sense) then yes. To access a field via any reflective getter or setter, you need to load it first, which when using `getDeclaredFields` or similar always requires choosing an element from the returned array.

OPAL/tac/src/main/scala/org/opalj/tac/fpcf/analyses/fieldaccess/reflection/ReflectionRelatedFieldAccessesAnalysis.scala
maximilianruesch: Fun fact: Its counterpart (in the set analysis) also does exactly!

OPAL/tac/src/main/scala/org/opalj/tac/fpcf/analyses/fieldaccess/reflection/ReflectionRelatedFieldAccessesAnalysis.scala
maximilianruesch: This continuation was for "for name classes", which got refactored in the meantime. I will (once again) copy them from `ReflectionRelatedCalls`, if we agree on this in person. I need some more background information here.

OPAL/tac/src/main/scala/org/opalj/tac/fpcf/analyses/fieldaccess/reflection/ReflectionRelatedFieldAccessesAnalysis.scala
maximilianruesch: See your other comment where you also asked which purpose this continuation has.

OPAL/tac/src/main/scala/org/opalj/tac/fpcf/analyses/fieldaccess/reflection/ReflectionRelatedFieldAccessesAnalysis.scala
maximilianruesch: I am against this, as it deteriorates two things:
1. It removes the hard typing support that every explicitly given type must be a `BaseType`
2. It makes us add a `UBTypeBasedFieldMatcher` always instead of just when a type is explicitly given

Point 2 can be workarounded (not really asthetically pleasing but its okay), but trying to fix Point 1 would necessitate usage like the `Either` Type which is again verbose and probably not what you want. Dropping Type support is also not what we want.

OPAL/tac/src/main/scala/org/opalj/tac/fpcf/analyses/fieldaccess/reflection/FieldMatcher.scala
maximilianruesch: A simple way to include inherited fields here would be to use the `DeclaredFields` here and check the "available" fields declared on the class, i.e. the fields declared on the class or its supertypes. However, this could only be done for defined fields, as we would need information on it being private or not (as private fields cannot be inherited).

Also, when the matcher is given a class and all its subtypes as the possible classes, we would compute the available fields many times over and over for the same classes. That is also included in the TODO one above.

Lets talk about this and the other comment shortly in person.

OPAL/tac/src/main/scala/org/opalj/tac/fpcf/analyses/fieldaccess/reflection/ReflectionRelatedFieldAccessesAnalysis.scala
errt: I don't understand what you are saying here. Why would removing the "None" here and putting it explicitly in the one place that uses this constructor without an explicit third parameter (i.e., line 981) have these consequences?

maximilianruesch: Apparently i misunderstood... I thought you wanted to give the types directly without wrapping in an `Option`. By "the default" i thought you referred to the `ObjectType.Object` down below given as the default for the API method return type, as opposed to the "None". Will implement.

OPAL/tac/src/main/scala/org/opalj/tac/fpcf/analyses/fieldaccess/reflection/FieldMatcher.scala
maximilianruesch: As discussed in person, i will open a separate PR where we address inherited fields properly, similar to `project.instanceMethods`.

OPAL/tac/src/main/scala/org/opalj/tac/fpcf/analyses/fieldaccess/reflection/ReflectionRelatedFieldAccessesAnalysis.scala
maximilianruesch: As discussed in person, we will migrate this to the `ClassForName` property as happened in reflective calls already.

OPAL/tac/src/main/scala/org/opalj/tac/fpcf/analyses/fieldaccess/reflection/ReflectionRelatedFieldAccessesAnalysis.scala
maximilianruesch: As discussed in person, we will move handling of such fields to a seperate methods that is called when an array load is encountered.

OPAL/tac/src/main/scala/org/opalj/tac/fpcf/analyses/fieldaccess/reflection/FieldMatcher.scala
maximilianruesch: Should be obsolete as per https://github.com/opalj/opal/pull/145#discussion_r1381668717. We can discuss this in more depth in the follow-up PR.

OPAL/tac/src/main/scala/org/opalj/tac/fpcf/analyses/fieldaccess/reflection/ReflectionRelatedFieldAccessesAnalysis.scala
```diff
@@ -362,15 +345,18 @@ sealed trait FieldInstanceBasedReflectiveFieldAccessAnalysis extends ReflectionA
...
-                if (!matchers.contains(NoFieldsMatcher))
-                    matchers += constructClassBasedFieldMatcherForClassVar(receiver.asVar, matchers, !isGetField)
-
-            case call @ VirtualFunctionCall(_, ObjectType.Class, _, "getDeclaredFields" | "getFields", _, receiver, _) =>
-                val isGetFields = call.name == "getFields"
-                if (isGetFields)
-                    matchers += PublicFieldMatcher
-                if (!matchers.contains(NoFieldsMatcher))
-                    matchers += constructClassBasedFieldMatcherForClassVar(receiver.asVar, matchers, !isGetFields)
+                if (!matchers.contains(NoFieldsMatcher)) {
```
errt, line 347: Should this be extracted to a new method to reuse in handleFieldArray?

maximilianruesch: Well in the old extracted function (which got removed again when splitting this function into `handleGetField` and `handleFieldArray`), we could use the surrounding information and did not have to pass everything into the function itself. 

Imo pulling this out into a function call is not worth, as it barely reduces the amount of parameters we have to pass.

errt: I'm less concerned with saving space than with ensuring that everything is handled consistently and that future changes don't break the analysis because one changed one location but not the other. That being said, I'm not sure whether extracting this to a method would actually help, thus the question. If you don't see a benefit in this, I'm fine with it.

DEVELOPING_OPAL/demos/src/main/scala/org/opalj/fpcf/analyses/FieldAccessInformationAnalysisDemo.scala
```diff
@@ -54,19 +60,25 @@ object FieldAccessInformationAnalysisDemo extends ProjectAnalysisApplication {
             case Some(requirements) => requirements + domain
         }
 
-        var propertyStore: PropertyStore = null
+        val propertyStore: PropertyStore = project.get(PropertyStoreKey)
         var analysisTime: Seconds = Seconds.None
         val analysesManager = project.get(FPCFAnalysesManagerKey)
+        val typeIterator = XTACallGraphKey.getTypeIterator(project)
+        project.updateProjectInformationKeyInitializationData(TypeIteratorKey) { _ => typeIterator }
+        project.updateProjectInformationKeyInitializationData(ContextProviderKey) { _ => typeIterator }
```
errt, line 66: I don't think it's necessary to configure both keys anymore. In fact, I think configuring the TypeIteratorKey produces an error now.

DEVELOPING_OPAL/demos/src/main/scala/org/opalj/fpcf/analyses/FieldAccessInformationAnalysisDemo.scala
```diff
@@ -54,19 +60,25 @@ object FieldAccessInformationAnalysisDemo extends ProjectAnalysisApplication {
...
+                        ++ Set(
+                            EagerFieldAccessInformationAnalysis,
+                            ReflectionRelatedFieldAccessesAnalysisScheduler
+                        )
                 )
-                ._1
             propertyStore.waitOnPhaseCompletion()
         } { t =>
-            analysisTime = t.toSeconds
+            analysisTime += t.toSeconds
```
errt: Why?

OPAL/tac/src/main/scala/org/opalj/tac/fpcf/analyses/fieldaccess/reflection/FieldMatcher.scala
```diff
@@ -0,0 +1,158 @@
...
+package fieldaccess
+package reflection
+
+import org.opalj.br.ObjectType
+import org.opalj.br.analyses.ProjectIndexKey
+import org.opalj.br.analyses.SomeProject
+import org.opalj.br.Field
+import org.opalj.br.FieldType
+import org.opalj.tac.fpcf.analyses.cg.V
+import org.opalj.tac.fpcf.analyses.cg.isTypeCompatible
```
errt: We should probably move these two out of the cg package

errt (changes requested):

DEVELOPING_OPAL/demos/src/main/scala/org/opalj/fpcf/analyses/FieldAccessInformationAnalysisDemo.scala
maximilianruesch: Probably a residual from a time i ran two phases of computation

DEVELOPING_OPAL/demos/src/main/scala/org/opalj/fpcf/analyses/FieldAccessInformationAnalysisDemo.scala
maximilianruesch: image

Apparently it does

OPAL/tac/src/main/scala/org/opalj/tac/fpcf/analyses/fieldaccess/reflection/FieldMatcher.scala
maximilianruesch: Imo we can move `V` to the `tac` package, as it is a common type of values used everywhere.

`isTypeCompatible` could be moved to `analysis` when only thinking about where it is used. However, one of the methods has a parameter `forCallReceiver` which toggles some behaviour inside the method and tightens type guarantees. The naming of this parameter suggests that is has something to do with calls which is rather specific to the `cg` package. I suggest renaming this parameter to `requireNonNullReferenceType` which is really what it enforces. Then we could also move this method to the `tac` package as it can be used generally in any setting there.

Thoughts?

errt: Sounds reasonable. Would move isTypeCompatible to the analyses package for now, can always move it one up if needed. Maybe with a shorter name for the parameter. `requireNonNull` should suffice

DEVELOPING_OPAL/demos/src/main/scala/org/opalj/fpcf/analyses/FieldAccessInformationAnalysisDemo.scala
errt: Now there's an unused import

maximilianruesch: WIP

OPAL/tac/src/main/scala/org/opalj/tac/fpcf/analyses/fieldaccess/reflection/FieldMatcher.scala
maximilianruesch: I disagree. We also require that the value is not a primitive type further down the method (when using the parameter).

errt: True, I only noticed the first occurence before.

errt (approved):
👍

PR #151, opened 2023-10-24
Fix return handling in TACNaive gen

kevzhumba:
The return handling for TAC naive always generated a return expression with a variable with ID 0, since that is supposed to represent the bottom element of the operand stack. However, the invariant in which variables are represented by their position in the operand stack is broken for instructions such as dup, which in its current state doesn't do anything to update the variable ids based on the new stack positions. Since at a return instruction the operand stack is guaranteed to have a single element, this change manually gets the variable at the top of the stack rather than defaulting to the variable with id 0.

Ran the unit tests and integration tests for TACNaive.


errt:
Looks good to me. Would it be sensible to instead change the behavior of `dup` and similar instructions?

kevzhumba:
I think that'd be a better solution for the long run and follows the intention of the original code (I see that the behavior for `swap` is correct and changes the variable identifiers). It would introduce a lot of additional assignment instructions into the generated TAC, which is why I just did this band-aid fix for now. Let me know what you think.

errt:
The comment in line 76 even says so, but it seems it has never been implemented like that. Having lots of additional assignments doesn't sound too good and it looks like our other TAC generation in class `TACAI` also just adds `NOP`s for `dup` instructions, but that has a comment on why it's permissible there (it also does `NOP` for `swap`). 

For now, I tend to go with your band-aid. However, I'm not deep enough into the TAC generation currently to really have an informed opinion. If you want to, you can try the other solution so we can have a look at that. In any case, please use `sbt scalariformFormat` so the CI accepts your code.

errt (approved):
LGTM

Issue #87, opened 2022-10-31
Fix Class.forName Handling in AI


TorunR:
Currently, Abstract Interpretation of the JDK fails in it:test.
This happens since there are new private versions of Class.forName since [this commit](https://github.com/openjdk/jdk/commit/c6339cb8a255d387bb182ad20dd69f3d460cf1ed). 
The error occurs in org/opalj/ai/domain/l1/ClassValues.scala in 
```
methodDescriptor match {
                        case `forName_String`                     => simpleClassForNameCall(pc, value)
                        case `forName_String_boolean_ClassLoader` => simpleClassForNameCall(pc, value)
                        case _ =>
                            throw new DomainException(
                                s"unsupported Class { ${methodDescriptor.toJava("forName")} }"
                            )
                    }
```
since we miss the branches for these new private Methods (and the public Method that uses a Module parameter).

errt:
Actually this does not fail for the third public method becaue that does not have a String as its first parameter. It should still probably be supported. Note though that it behaves differently compared to the other two methods (e.g., it returns null instead of throwing ClassNotFoundException), so simply invoking simpleClassForNameCall won't suffice.

TorunR:
Okay so this could be 2 Issues I guess. 
We should check whether `Reflection.getCallerClass()` is handled correctly as well.

errt:
The exception is now fixed with #97. The issues of more Class.forName methods to be handled remains open.

johannesduesing:
As discussed in the last OPAL call, i'll have a look into the remainder of this issue.

johannesduesing:
Closed by #152 and #153

PR #152, opened 2023-11-03
Support Class.forName(Module, String) in abstract interpretation

johannesduesing:
**Reason for this PR**
This PR is to support tracking of Class values in abstract interpretation when `Class.forName(Module, String)` or `Class.forName(Module, String, Class)` are invoked, see #87.

**Changes in this PR**
`org.opalj.ai.domain.l1.ClassValues` now explicitly handles calls to the new `Class.forName` methods involving Modules introduced in Java 9.  The actual behavior of these methods is as follows: If the class with the given name is part of the given Module, then the class is returned, otherwise `null` is returned. For the abstract interpretation we overapproxmiate this behavior and always return the class with the given name. This is because currently there is no tracking of values of type `Module`, which would allow us to actually check whether the class belongs to the given module or not.

**Note**
Testing this implementation requires me to compile the test fixtures `ai.jar` with Java 9 instead of Java 8 (because `Class.forName(Module, String)` was introduced in Java 9). This lead to the fixture jar being renamed to `ai-9.jar` (i did not find a fixture compiler option to specify the output filename) which in turn caused a lot of test files being changed, since i had to replace `"ai.jar"` with `"ai-9.jar"`.


OPAL/ai/src/main/scala/org/opalj/ai/domain/l1/ClassValues.scala
```diff
@@ -184,8 +189,23 @@ trait ClassValues
                             )
                     }
 
+                case _ if secondOperandIsStringValue =>
```
errt: I wonder if this matching could be done in a nicer way, maybe along the lines of
```
operands match {
  case _ :+ (sv: StringValue) =>
  case _ :+ (sv: StringValue) :+ _ =>
}
```
?

errt: Or maybe do away with the value match and just match on the descriptor?

errt (dismissed):

errt:
Renamed tests are okay, another option would have been to separate out the new tests into their own fixture.

OPAL/ai/src/main/scala/org/opalj/ai/domain/l1/ClassValues.scala
johannesduesing: I now changed this so it uses `operands.reverse match { ... }` - should essentially be the same.

errt: It creates a reversed list, though. Probably not a performance issue here, though.

johannesduesing: Ah yes that's right - i changed that to your suggestion. Thought the reversed version would be slightly more readable, as the first operand would actually be the first in the match clauses, but i think performance should be more important here. Could you re-approve your review so i can merge?

errt: Did you make sure it works? Readability can be enhanced by commenting on what the cases do.

errt (approved):

OPAL/ai/src/main/scala/org/opalj/ai/domain/l1/ClassValues.scala
johannesduesing: Well i ran all the new unit tests and it does yield the desired results - it seems to be functionally equivalent to the previous solutions. Comments for readability are already in place, so i think we're fine there.

PR #156, opened 2023-11-14
Address some deprecation warnings in build definition

johannesduesing:
This PR addresses some of the deprecation warnings the showed up in SBT when execution `cleanBuild` (and other tasks). In particular, the following issues have been dealt with:

* `warning: lazy value itSettings in object Defaults is deprecated (since 1.9.0): [..]`
* `[warn] Ignored unknown package option FixedTimestamp`
* `[warn] sbt 0.13 shell syntax is deprecated; use slash syntax instead`

### DIFF
--- a/.github/workflows/run_it_test_on_comment.yml
+++ b/.github/workflows/run_it_test_on_comment.yml
@@ -54,9 +54,9 @@ jobs:
           status: pending
 
       - name: Run Integration Tests
-        run: sbt -J-Xmx40G cleanBuild it:test
+        run: sbt -J-Xmx40G "cleanBuild; IntegrationTest / test"
       - name: Run Cross Language Tests
-        run: sbt -J-Xmx40G "project LLVM;it:test;project ValidateCross;it:test"
+        run: sbt -J-Xmx40G "project LLVM; IntegrationTest / test;project ValidateCross; IntegrationTest / test"
         
       - name: Set latest commit status as ${{ job.status }}
         uses: myrotvorets/set-commit-status-action@655d7d2517bab7f5d1b6e74cd7d5e995264184b1
--- a/build.sbt
+++ b/build.sbt
@@ -82,15 +82,15 @@ ThisBuild / javaOptions ++= Seq(
 addCommandAlias(
   "compileAll",
   "; copyResources ; scalastyle ; " +
-    "test:compile ; test:scalastyle ; " +
-    "it:scalariformFormat ; it:scalastyle ; it:compile "
+    "OPAL / test / compile ; test:scalastyle ; " +
+    "OPAL / IntegrationTest / scalariformFormat ; OPAL / IntegrationTest / scalastyle ; OPAL / IntegrationTest / compile "
 )
 
 addCommandAlias("buildAll", "; compileAll ; unidoc ;  publishLocal ")
 
 addCommandAlias(
   "cleanAll",
-  "; clean ; cleanCache ; cleanLocal ; test:clean ; it:clean ; cleanFiles"
+  "; clean ; cleanCache ; cleanLocal ; OPAL / Test / clean ; OPAL / IntegrationTest / clean ; cleanFiles"
 )
 
 addCommandAlias("cleanBuild", "; project OPAL ; cleanAll ; buildAll ")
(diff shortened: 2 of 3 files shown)


errt (dismissed):
Looks good to me

johannesduesing:
Weird things are happening with the integration test, git fails to checkout the current commit and exists with code `1`.
```
/usr/bin/git -c protocol.version=2 fetch --no-tags --prune --progress --no-recurse-submodules --depth=1 origin +refs/heads/feature/sbt-deprecation-warnings*:refs/remotes/origin/feature/sbt-deprecation-warnings* +refs/tags/feature/sbt-deprecation-warnings*:refs/tags/feature/sbt-deprecation-warnings*
The process '/usr/bin/git' failed with exit code 1
Waiting 14 seconds before trying again
/usr/bin/git -c protocol.version=2 fetch --no-tags --prune --progress --no-recurse-submodules --depth=1 origin +refs/heads/feature/sbt-deprecation-warnings*:refs/remotes/origin/feature/sbt-deprecation-warnings* +refs/tags/feature/sbt-deprecation-warnings*:refs/tags/feature/sbt-deprecation-warnings*
Error: The process '/usr/bin/git' failed with exit code 1
```

I've re-run this three times with the same result ... I'll retry again tomorrow, maybe there's an issue with GitHub and / or the runner. If you've seen this before @errt please let me know.

errt:
Weird. I haven't seen this before. Maybe a restart of the runner could help. @TorunR ?

TorunR:
A restart won't help. The problem is that the runner tries to check out the branch from johannes' fork. I think we have to fix that in the action somehow. Learnt that here: (https://github.com/WeblateOrg/weblate/issues/6240#issuecomment-869715886)
I don't have time to look into this further. sorry

TorunR:
https://github.com/opalj/opal/actions/runs/6874395149/job/18695980061#step:3:47
Yes https://github.com/opalj/opal/pull/157 did not help

johannesduesing:
That's too bad .. i'll keep looking for a better solution - if there is none i'll at least revert to the old one since the current state doesn't work at all

TorunR:
I see two options: Either find out whether you can find out the PR ref that is on origin so that it looks like a regular PR Action: [like this](https://github.com/opalj/opal/actions/runs/6865488949/job/18669611615#step:2:40)
Or use the repository argument with the checkout:
```
- uses: actions/checkout@v4
  with:
    # Repository name with owner. For example, actions/checkout
    # Default: ${{ github.repository }}
    repository: '
```

TorunR:
While we are changing the actions anyway, do we need to use the slash syntax there as well?

TorunR:
https://github.com/opalj/opal/actions/runs/6874896394/job/18697425720#step:3:495
still develop

johannesduesing:
Hm very odd, the arguments don't even seem to make it to the actual [Action invocation](https://github.com/opalj/opal/actions/runs/6874896394/job/18697425720#step:3:12) - i'll keep looking for a fix

TorunR:
I think the arguments make it there, but the comment stuff is run on develop. That's the reason we use the action in the first step, to get the ref to the pr branch.
EDIT: Oh, I think you are right. Maybe we need to update action?

johannesduesing:
I'll try updating to v4 later today, maybe that helps. Only other thing i can imagine is that the interpolation with `${{  }}` is wrong for github API calls .. in line 11 we use `if github.event.issue.pull_request` without any interpolation syntax, but i am not confident that is the same thing.

EDIT: Nvm i think i know what the problem is. The action triggeres on `issue_comment`, which provides an API interface that only has very limited PR information available (via `github.event.issue.pull_request`), [see here](https://docs.github.com/de/webhooks/webhook-events-and-payloads#issue_comment). What i was looking at was the API interface for `pull_request` which is not available here. I am not confident that it is possible to do what we plan this way, but i'll try some more this afternoon

TorunR:
Maybe we should remove the comment stuff and figure out whether https://docs.github.com/en/actions/using-workflows/manually-running-a-workflow using the UI works better

johannesduesing:
Sounds like a good plan - let's discuss that in tomorrows meeting maybe? I'll test some more on a dedicated repo, so i don't have to spam additional PRs here anymore.

johannesduesing:
@TorunR The issue seems to be resolved, finally. [The Logs](https://github.com/opalj/opal/actions/runs/6891834652/job/18747780675#step:4:17) show that this time the correct HEAD has been checked out, and setting the commit status seems to be working as it did before. I'll let this run finish so the PR can be merged.

TorunR (dismissed):

TorunR:
Thank you for this. Sorry that this was way harder than anticipated.

I still have this question:
> While we are changing the actions anyway, do we need to use the slash syntax there as well?

johannesduesing:
Ah right sorry i forgot about that. Sure we can do that, do you want that to be changed only in the actual call inside the workflow, or also for the magic phrase itself?

TorunR:
It think the workflows are enough

TorunR (approved):

PR #163, opened 2023-12-01
Fix critical field access information problems

maximilianruesch:
Fixes two problems from the field access information analysis PRs:
1. From #126, the state handling for the continuation in the `AbstractPointsToAnalysis` was faulty and led to inconsistent states
2. From #145, the getter and setter methods from MethodHandle lookup objects needed adding to the reference.conf


errt (approved):
LGTM

PR #172, opened 2023-12-21
Scalariform removal precursor

errt:
This PR contains some uncontroversial changes as a preparation to the Scalariform removal and to keep changes in the main PR #122  smaller and easier to review.

This is mostly import order, spaces around string concatenation and a few minor whitespace changes.

First commit is what I'd like to have, second is what is necessary to get it past Scalariform as long as we still have that on.

The third commit shows an important issue: Naive lexical ordering of imports breaks code.


TorunR (approved):
Didn't check every file. One idea would be to fix the license headers on the java files

errt:
Yes, Java file headers should probably be fixed, too.

PR #174, opened 2023-12-22
Fix ClassImmutabilityAnalysis

errt:
How did it end up so broken?

### DIFF
--- a/OPAL/br/src/main/scala/org/opalj/br/fpcf/analyses/immutability/ClassImmutabilityAnalysis.scala
+++ b/OPAL/br/src/main/scala/org/opalj/br/fpcf/analyses/immutability/ClassImmutabilityAnalysis.scala
@@ -8,11 +8,7 @@ package immutability
 import scala.collection.immutable.SortedSet
 
 import org.opalj.br.ClassFile
-import org.opalj.br.ClassSignature
-import org.opalj.br.ClassTypeSignature
-import org.opalj.br.FormalTypeParameter
 import org.opalj.br.ObjectType
-import org.opalj.br.SimpleClassTypeSignature
 import org.opalj.br.analyses.ProjectInformationKeys
 import org.opalj.br.analyses.SomeProject
 import org.opalj.br.fpcf.FPCFAnalysis
@@ -29,11 +25,11 @@ import org.opalj.br.fpcf.properties.immutability.NonTransitivelyImmutableClass
 import org.opalj.br.fpcf.properties.immutability.NonTransitivelyImmutableField
 import org.opalj.br.fpcf.properties.immutability.TransitivelyImmutableClass
 import org.opalj.br.fpcf.properties.immutability.TransitivelyImmutableField
-import org.opalj.fpcf.ELBP
 import org.opalj.fpcf.Entity
 import org.opalj.fpcf.EOptionP
 import org.opalj.fpcf.EPK
 import org.opalj.fpcf.EPS
+import org.opalj.fpcf.EUBPS
 import org.opalj.fpcf.FinalEP
 import org.opalj.fpcf.FinalP
 import org.opalj.fpcf.IncrementalResult
@@ -106,8 +102,8 @@ class ClassImmutabilityAnalysis(val project: SomeProject) extends FPCFAnalysis {
         var results: List[ProperPropertyComputationResult] = List(result)
         var nextComputations: List[(PropertyComputation[ClassFile], ClassFile)] = Nil
         val directSubtypes = classHierarchy.directSubtypesOf(t)
-        directSubtypes.foreach { t =>
-            project.classFile(t) match {
+        directSubtypes.foreach { subtype =>
+            project.classFile(subtype) match {
                 case Some(scf) =>
                     nextComputations ::= (
                         (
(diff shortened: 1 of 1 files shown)


OPAL/br/src/main/scala/org/opalj/br/fpcf/analyses/immutability/ClassImmutabilityAnalysis.scala
```diff
@@ -118,38 +114,18 @@ class ClassImmutabilityAnalysis(val project: SomeProject) extends FPCFAnalysis {
...
+                        s"missing class file of ${subtype.toJava}; setting all subtypes to mutable"
                     )
-                    results ::= createResultForAllSubtypes(t, MutableClass)
+                    results ::= createResultForAllSubtypes(subtype, MutableClass)
             }
         }
         IncrementalResult(Results(results), nextComputations.iterator)
     }
 
-    def determineGenericTypeBounds(classFile: ClassFile): Set[(String, String)] = {
```
maximilianruesch, line 129: Question: This is just an unused function that somehow ended up as a residual?

OPAL/br/src/main/scala/org/opalj/br/fpcf/analyses/immutability/ClassImmutabilityAnalysis.scala
```diff
@@ -350,33 +318,30 @@ class ClassImmutabilityAnalysis(val project: SomeProject) extends FPCFAnalysis {
...
-                case UBP(MutableField)    => return Result(t, MutableClass);
-                case ELBP(e, NonTransitivelyImmutableField |
-                    TransitivelyImmutableField) => dependees -= e
-                case UBP(TransitivelyImmutableField)    => // no information about field mutability
-                case UBP(NonTransitivelyImmutableField) => maxLocalImmutability = NonTransitivelyImmutableClass
+
+                case FinalP(MutableField)                                            => return Result(t, MutableClass);
+                case UBP(MutableField)                                               => return Result(t, MutableClass);
+
+                case LBP(NonTransitivelyImmutableField | TransitivelyImmutableField) =>
```
maximilianruesch: This case now does not have an effect and thus needs a comment explaining why it does nothing (just like below)

OPAL/br/src/main/scala/org/opalj/br/fpcf/analyses/immutability/ClassImmutabilityAnalysis.scala
```diff
@@ -350,33 +318,30 @@ class ClassImmutabilityAnalysis(val project: SomeProject) extends FPCFAnalysis {
...
+                case UBP(NonTransitivelyImmutableField)                              => maxLocalImmutability = NonTransitivelyImmutableClass
+
                 case UBP(DependentlyImmutableField(parameter)) if maxLocalImmutability != NonTransitivelyImmutableClass =>
                     genericTypeParameters ++= parameter
-                    maxLocalImmutability =
-                        DependentlyImmutableClass(genericTypeParameters)
-                case _ => Result(t, MutableClass) // TODO
+                    maxLocalImmutability = DependentlyImmutableClass(genericTypeParameters)
+
+                case _ => Result(t, MutableClass) //TODO
```
maximilianruesch: What is "TODO" here?

OPAL/br/src/main/scala/org/opalj/br/fpcf/analyses/immutability/ClassImmutabilityAnalysis.scala
```diff
@@ -193,12 +169,10 @@ class ClassImmutabilityAnalysis(val project: SomeProject) extends FPCFAnalysis {
         }
     }
 
-    private[this] object SuperClassKey
```
maximilianruesch, line 196: Question to check if understand this correctly: Using this singleton previously only allowed one type of dependency on super class information?

maximilianruesch (changes requested):
Some comments, some questions. Maybe we could scan the remaining code for "[DEBUG]" comments sometime...

OPAL/br/src/main/scala/org/opalj/br/fpcf/analyses/immutability/ClassImmutabilityAnalysis.scala
errt: Looks like it. Don't know why.

OPAL/br/src/main/scala/org/opalj/br/fpcf/analyses/immutability/ClassImmutabilityAnalysis.scala
errt: I think so, but it seems if that was ever really necessary, it was in an earlier version with a different analysis design. When I removed it, the analysis broke, but only because wrong dependencies were registered.

OPAL/br/src/main/scala/org/opalj/br/fpcf/analyses/immutability/ClassImmutabilityAnalysis.scala
errt: I wondered as well...

maximilianruesch: I looked into its history. It once was "TODO check", which is only that tad more meaningful. One of your colleagues removed and readded this TODO (the latter in a reversion commit).

Given it is very old and not meaningful, i think we can remove it.

maximilianruesch (approved):
LGTM. Lets wait what @roterEmil has to say to this.

roterEmil (approved):

PR #1, opened 2021-04-12
String Analysis

errt:
Implementation of the string analysis (the interprocedural as well as the intraprocedural variant).

Original PR by PatrickN.

PR #178, opened 2024-01-08
Add test setup for alias analysis

LukasKlenner:
This PR adds a setup for testing alias analyses, including some (simple) test cases.

More (complex) test cases alongside analyses that pass them will be added in future PR's.
This is a continuation of #171


errt:
Build doesn't work for several reasons, please fix

LukasKlenner:
building issues are fixed, but it fails because the AliasTests don't pass, despite them being annotated with @Ignored. Is it not supported or is something else required to make the tests not being executed?

LukasKlenner:
PS: AliasEntity and AliasSourceElement need to be in the tac package (and not br) because they rely on the DefinitionSite class

DEVELOPING_OPAL/validate/src/test/java/org/opalj/fpcf/properties/alias/Alias.java
```diff
@@ -0,0 +1,45 @@
+package org.opalj.fpcf.properties.alias;
+
+import java.lang.annotation.*;
```
errt: Do not use multi-imports

DEVELOPING_OPAL/validate/src/test/java/org/opalj/fpcf/fixtures/alias/ShapeTest.java
```diff
@@ -0,0 +1,229 @@
...
+import org.opalj.fpcf.properties.alias.NoAlias;
+
+public class ShapeTest {
+
+    public static void main(String[] args) {
+        ShapeTest shapeTest = new ShapeTest();
+        Shape a = shapeTest.addRectangle(createCircle(1.0));
+
+        if (a != null) {
+            System.out.println("circle added");
```
errt: What is the point of these checks and printlns?

DEVELOPING_OPAL/validate/src/test/java/org/opalj/fpcf/fixtures/alias/ShapeTest.java
```diff
@@ -0,0 +1,229 @@
...
+    }
+
+    @Alias(mayAlias = {@MayAlias(reason = "param might be added to field", id = "ShapeTest.shapesAddParam")},
+            noAlias = {@NoAlias(reason = "param is never added to field", id = "ShapeTest.shapeGetLastOrParam")})
+    Shape[] shapes = new Shape[10];
+    int index = 0;
+
+    Shape currentShape;
+    
+    @Alias(mayAlias = {@MayAlias(reason = "", id = "ShapeTest.addRectangle")})
```
errt: Reasons missing

DEVELOPING_OPAL/validate/src/test/java/org/opalj/fpcf/fixtures/alias/ShapeTest.java
```diff
@@ -0,0 +1,229 @@
...
+
+        if (nullShape == null) {
+            System.out.println("created null");
+        }
+
+        shapeTest.getLastOr(new Square(1.0));
+        shapeTest.getLastOrCircle(new Square(1.0));
+    }
+
+    @Alias(mayAlias = {@MayAlias(reason = "param might be added to field", id = "ShapeTest.shapesAddParam")},
```
errt: I think you should use an @Repeatable annotation to avoid these strange nested annotations. Have a look at properties/callgraph/DirectCall and properties/callgraph/DirectCalls for an example.

DEVELOPING_OPAL/validate/src/test/java/org/opalj/fpcf/fixtures/alias/ShapeTest.java
```diff
@@ -0,0 +1,229 @@
...
+            @MayAlias(reason = "return value might be null", id = "ShapeTest.getLastOrNull", aliasWithNull = true)})
+    public Shape getLastOrCircle(
+            @Alias(mayAlias = {@MayAlias(reason = "return value might be param", id = "ShapeTest.getLastOrCircleParam")},
+                   noAlias = {@NoAlias(reason = "param is never new circle", id = "ShapeTest.getLastOrCircleParamNewCircle")}
+            )
+            Shape shape) {
+
+        Shape currentShape = shape;
+
+        Shape circle = new @Alias(mayAlias = {@MayAlias(reason = "return value might be new circle", id = "ShapeTest.getLastOrCircleNewCircle")},
```
errt: Shouldn't the annotation be on the variable, not the constructor call?

DEVELOPING_OPAL/validate/src/test/java/org/opalj/fpcf/fixtures/alias/ShapeTest.java
errt: Does this annotate the array variable or the array contents? There should be a clear distinction between the two

DEVELOPING_OPAL/validate/src/test/java/org/opalj/fpcf/fixtures/alias/ShapeTest.java
```diff
@@ -0,0 +1,229 @@
...
+
+        for (int i = 0; i < shapes.length; i++) {
+
+            Shape next = shapes[i];
+            if (next != null) {
+                currentShape = next;
+            }
+        }
+
+        if (!(currentShape instanceof Circle)) {
```
errt: I'm not sure what the point in all of these checks is. Try to write the tests so that for every line that is there, it is clear why it is necessary and what it is supposed to test. Try to separate different things that you want to test into different files with each testing one specific category of things. Name variables, functions, etc. in a way that describes what they test. Rectangles, shapes, adding them, etc. doesn't have a meaning in terms of the test.

DEVELOPING_OPAL/validate/src/test/scala/org/opalj/fpcf/properties/alias/AliasPropertyMatcher.scala
```diff
@@ -0,0 +1,68 @@
...
+        entity:     Any,
+        a:          AnnotationLike,
+        properties: Iterable[Property]
+    ): Option[String] = {
+
+        if (properties.isEmpty) {
+            return Some("No Properties assigned to entity")
+        }
+        if (properties.count(_.isInstanceOf[Alias]) > 1) {
+            return Some("Multiple alias properties found")
```
errt: Can that even happen?

DEVELOPING_OPAL/validate/src/test/scala/org/opalj/fpcf/properties/alias/AliasPropertyMatcher.scala
```diff
@@ -0,0 +1,68 @@
...
+    override def validateProperty(
+        p:          Project[_],
+        as:         Set[ObjectType],
+        entity:     Any,
+        a:          AnnotationLike,
+        properties: Iterable[Property]
+    ): Option[String] = {
+
+        if (properties.isEmpty) {
+            return Some("No Properties assigned to entity")
```
errt: Your `isRelevant` method should probably prevent this from happening.

DEVELOPING_OPAL/validate/src/test/java/org/opalj/fpcf/fixtures/alias/ShapeTest.java
```diff
@@ -0,0 +1,229 @@
+package org.opalj.fpcf.fixtures.alias;
+
+import org.opalj.fpcf.properties.alias.Alias;
+import org.opalj.fpcf.properties.alias.MayAlias;
+import org.opalj.fpcf.properties.alias.MustAlias;
+import org.opalj.fpcf.properties.alias.NoAlias;
+
+public class ShapeTest {
```
errt: Wouldn't call these things "Test", they aren't tests, they are fixtures used by tests. Naming convention in OPAL here is just to name them after what they contain.

errt (changes requested):

errt:
> PS: AliasEntity and AliasSourceElement need to be in the tac package (and not br) because they rely on the DefinitionSite class

This may hint at potential to improve. In many cases, we map TAC use sites to bytecode program counters and values (search for `persistentUVar`), a similar mapping, maybe even excluding the values, might be applicable here.

LukasKlenner:
All previous issues have been addressed, but most of the classes have been completely reworked

DEVELOPING_OPAL/validate/src/test/java/org/opalj/fpcf/fixtures/alias/ParameterAlias.java
```diff
@@ -0,0 +1,97 @@
...
+
+        pa.mayAliasThisParamTwoMethods1();
+        pa.mayAliasThisParamTwoMethods2();
+
+        pa.noAliasThisParamTwoMethods1();
+        pa2.noAliasThisParamTwoMethods2();
+    }
+
+    public static void noAliasWithLocal(@NoAliasLine(reason = "noAlias with uVar", lineNumber = 44) Object o1) {
+
```
errt: Remove these blank lines

DEVELOPING_OPAL/validate/src/test/java/org/opalj/fpcf/fixtures/alias/StaticFieldAlias.java
```diff
@@ -0,0 +1,92 @@
...
+    }
+
+    @MayAlias(reason = "may alias with static final field and return value", id = 4)
+    public static Object returnStaticFinalField() {
+        Object o = staticFinalField;
+        return o;
+    }
+
+    @MayAlias(reason = "may alias with field and return value via parameter", id = 5)
+    @MayAliasLine(reason = "may alias with return value and uVar via parameter and field", lineNumber = 43, methodName = "reassignField")
```
errt: What is the reason to have @MayAliasLine in addition to @MayAlias? Shouldn't one suffice?

OPAL/br/src/main/scala/org/opalj/br/fpcf/properties/alias/AliasSourceElement.scala
```diff
@@ -0,0 +1,280 @@
...
+
+    override def asAliasFP: AliasFormalParameter = this
+}
+
+/**
+ * A persistent representation (using pcs instead of TAC value origins) for a UVar.
+ *
+ * @see [[org.opalj.tac.fpcf.analyses.cg.persistentUVar]]
+ */
+case class PersistentUVar(valueInformation: ValueInformation, defSites: IntTrieSet)
```
errt: This should not be part of this file, but be in a generic place. You might want to take the code from here:
https://github.com/opalj/opal/pull/1/files#diff-62df04527a62216c2330e018931c309e5a1b88dac0e66e6a11e451a89dcf0517
and the following files

DEVELOPING_OPAL/validate/src/test/java/org/opalj/fpcf/fixtures/alias/StaticFieldAlias.java
LukasKlenner: They are referencing two different entities. Line 76 states that the return value of the method is in a may alias relation with the field in general. Line 77 states that the return value of the method is in a may alias relation with the uVar that the field is assigned to in line 43.

OPAL/br/src/main/scala/org/opalj/br/fpcf/properties/alias/AliasSourceElement.scala
LukasKlenner: How should i handle that the other PR is not merged yet? Just copy&paste the relevant code and add it to this PR?

errt: Yes, that's probably the only practical solution for now. Unless that part of the code is in isolated commits (I didn't check that), then you could cherry-pick them.

DEVELOPING_OPAL/validate/src/test/java/org/opalj/fpcf/fixtures/alias/StaticFieldAlias.java
errt: I think my question was more: Why do we need two different types of annotations. Can't one type of annotation express both of these entities?

LukasKlenner: Oh, i see. Currently, the original MayAlias annotation is used if both element can be annotated, e.g., an alias between two parameters. MayAliasLine has been added to explicitly handle cases where at least one of the elements cannot be annotated (is an uVar). Technically, it should be possible to add the `int id();` method to the mayAliasLine annotation with a default value and choosing the resolution strategy (searching for annotation with the same id vs. searching for uVar(s) in given lineNumber) depending on whether id is explicitly set or not. That would eliminate the need for two types of annotations. I will look into that

LukasKlenner: There is now only one type of annotation

OPAL/br/src/main/scala/org/opalj/br/fpcf/properties/alias/AliasSourceElement.scala
LukasKlenner: Cherry-picking this commit (https://github.com/opalj/opal/commit/c3b9715575a51b44bd97184b8acd8fd69c29af7c) should be sufficient. However, the class has been created in the tac package, therefore i can't use it as it is. I could move it to the br package but i don't know if this is intended and that should be done in the other PR as well to avoid duplicates after merging both.

LukasKlenner: I have cherry picked the commit now and moved the class to the br package with some neccessary adjustments. As mentioned, this should probably also be done in the other PR (if you want to keep it this way)

errt (approved):
LGTM

DEVELOPING_OPAL/validate/src/test/java/org/opalj/fpcf/fixtures/alias/ArrayAlias.java
```diff
@@ -0,0 +1,122 @@
+/* BSD 2-Clause License - see OPAL/LICENSE for details. */
+package org.opalj.fpcf.fixtures.alias;
+
+import org.opalj.fpcf.properties.alias.MayAlias;
+import org.opalj.fpcf.properties.alias.NoAlias;
+//import org.opalj.tac.fpcf.analyses.alias.AllocationSiteBasedAliasAnalysis;
```
errt: Remove commented out import

OPAL/br/src/main/scala/org/opalj/br/PDUVar.scala
```diff
@@ -0,0 +1,86 @@
+/* BSD 2-Clause License - see OPAL/LICENSE for details. */
+package org.opalj.br
```
errt: Fix package declaration (move `br` to line of its own)

errt (changes requested):

errt (approved):

PR #182, opened 2024-01-15
Exchange Scalariform for Scalafmt

errt:
This is my proposal for the scalafmt formatting. Please comment on things that you think are formatted badly, but note that scalafmt is quite strict and most things probably can't be changed or would have other negative consequences elsewhere.

Primary review target should be the "Reformat everything" commit, everything else is smaller changes that are probably not worth going over them in detail.

One choice that we do have is the indentation of some expressions, I pushed the alternative version to another branch so you can have a look at the diff and comment on which one you like better: https://github.com/opalj/opal/compare/feature/scalariformRemovalPreparation...feature/scalafmtAlternative

I'm inclined to set indent.ctorSite (indentation of parameters in constructors) down to 4, but will do so in a follow-up PR to keep review load here lower, since our current scalariform formatting uses an indent of 8 here (at least in most places).


JuliusNmn (dismissed):
lgtm

TorunR (dismissed):

DEVELOPING_OPAL/demos/src/main/scala/org/opalj/ai/InfiniteRecursions.scala
```diff
@@ -139,27 +139,29 @@ object InfiniteRecursions extends ProjectAnalysisApplication {
...
+                                case v: domain.ALongValue =>
+                                    if (localsArray(0).exists(_ eq v))
+                                        true // the value is parameter
+                                    else
+                                        false
+                                case _: domain.LongSet      => true
+                                case _: domain.IntegerRange => true
+                                case _                      => false
+                            }
+                        )
```
maximilianruesch, line 164: In this case, the opening braces (on the if statement and forall) are on one line and the closing braces are on two different lines. Can this be made consistent?

maximilianruesch (dismissed):
Looked at about 5% of the files, should suffice.

On the alternative formatting: I kind of dislike the switcheroo between 
https://github.com/opalj/opal/compare/feature/scalariformRemovalPreparation...feature/scalafmtAlternative#diff-e7c81d919d3336a35a35e2fd58094e07ee59f8f647eb5614abcf5f387dd5fd8eR79-R81

and https://github.com/opalj/opal/compare/feature/scalariformRemovalPreparation...feature/scalafmtAlternative#diff-10e1883cf2a6be78cdd52602faf1a3ddf01a0b718ba68857a5e0fda48550cc7cR87-R88

Imo both should be indentet if possible. If not, i prefer the leftside version (where the changes i marked here are not made).

errt:
> Looked at about 5% of the files, should suffice.
> 
> On the alternative formatting: I kind of dislike the switcheroo between [feature/scalariformRemovalPreparation...feature/scalafmtAlternative#diff-e7c81d919d3336a35a35e2fd58094e07ee59f8f647eb5614abcf5f387dd5fd8eR79-R81](https://github.com/opalj/opal/compare/feature/scalariformRemovalPreparation...feature/scalafmtAlternative#diff-e7c81d919d3336a35a35e2fd58094e07ee59f8f647eb5614abcf5f387dd5fd8eR79-R81)
> 
> and [feature/scalariformRemovalPreparation...feature/scalafmtAlternative#diff-10e1883cf2a6be78cdd52602faf1a3ddf01a0b718ba68857a5e0fda48550cc7cR87-R88](https://github.com/opalj/opal/compare/feature/scalariformRemovalPreparation...feature/scalafmtAlternative#diff-10e1883cf2a6be78cdd52602faf1a3ddf01a0b718ba68857a5e0fda48550cc7cR87-R88)
> 
> Imo both should be indentet if possible. If not, i prefer the leftside version (where the changes i marked here are not made).

Interesting case, I have no idea why that is. The alternative should only have more indentation, not less, but looks like that's not always the case. Left is this PR, right?

DEVELOPING_OPAL/demos/src/main/scala/org/opalj/ai/InfiniteRecursions.scala
errt: Closing braces/parens on one line I don't think are possible. Opening ones on two lines might be possible by manually formatting them that way, scalafmt might not revert that.

roterEmil (approved):

PR #187, opened 2024-01-22
Scalafmt introduction post work

errt:
Some more changes in the wake of removing Scalariform in favor of Scalafmt


maximilianruesch:
We can probably address some concerns from https://github.com/opalj/opal/pull/182#pullrequestreview-1828147803 using the `indentOperator.excludeRegex` from https://scalameta.org/scalafmt/docs/configuration.html#indentoperator with an impossible regex like `(?!x)x` which would effectively evaluate `indentOperator.exemptScope` to `None`.

errt:
Not sure from reading the documentation, but isn't exemptScope stronger than excludeRegex? We can still try, it should maybe indent some boolean expressions better, as the default for excludeRegex is && and ||.

maximilianruesch:
Imo they can be made the same strength:
- `exemptScope`: "This parameter determines when an infix operator can be exempted from applying continuation indentation."
- `excludeRegex` "Defines a regular expression for excluded infix operators. If an eligible operator matches, it will not be indented."

When "continuation indentation" and "it will be indented" mean the same, setting an impossible regex should just emulate `exemptScope=None`.

Regardless of how strong it is, imo worth a try.

errt:
Works, but I put it in another PR: #188

JuliusNmn (approved):

PR #188, opened 2024-01-23
Additional operator indentation

errt:
As suggested by https://github.com/opalj/opal/pull/187#issuecomment-1904646058

Not sure I like it though, in some places it adds additional indentation to expressions that are already indented a lot.
E.g.:
https://github.com/opalj/opal/blob/40e1c0adaa29e541359d0941b7782d90b1c9e40f/OPAL/br/src/main/scala/org/opalj/br/analyses/AnalysisApplication.scala#L160C1-L166C56

### DIFF
--- a/.scalafmt.conf
+++ b/.scalafmt.conf
@@ -13,7 +13,11 @@ indent {
   main = 4
   callSite = 4
 }
-indentOperator.exemptScope = aloneEnclosed
+
+indentOperator {
+  excludeRegex = "^$"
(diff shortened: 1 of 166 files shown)


errt:
Note, please treat this as an RFC and comment on whether you are in favor of these changes or not.

maximilianruesch:
I am in favor. The code sample above makes sense, if its too indented in some cases there should always be the option to break the first condition to the next line and reduce indentation that way.

PR #191, opened 2024-02-02
Fix initialization of field access information key

maximilianruesch:
Fixes an initialization bug for the `FieldAccessInformationKey`. When no initialization data was configured, the requirements stated that as a fallback, the `SimpleFieldAccessInformationAnalysis` will be used.

Actually, the requirements only used this fallback for determining `ProjectInformationKey` dependencies but not for running the actual key itself. This subsequently caused the `compute` step to fail.

### DIFF
--- a/OPAL/br/src/main/scala/org/opalj/br/analyses/FieldAccessInformationKey.scala
+++ b/OPAL/br/src/main/scala/org/opalj/br/analyses/FieldAccessInformationKey.scala
@@ -19,19 +19,15 @@ import org.opalj.log.OPALLogger
 object FieldAccessInformationKey extends ProjectInformationKey[FieldAccessInformation, Seq[FPCFAnalysisScheduler]] {
 
     override def requirements(project: SomeProject): ProjectInformationKeys = {
-        val schedulers = project.getProjectInformationKeyInitializationData(this) match {
-            case Some(s) => s
-            case None =>
+        val schedulers = project.getOrCreateProjectInformationKeyInitializationData(
+            this, {
                 OPALLogger.warn(
                     "analysis configuration",
                     s"no field access information analysis configured, using SimpleFieldAccessInformationAnalysis as a fallback"
                 )(project.logContext)
-                project.updateProjectInformationKeyInitializationData(this) {
-                    case None             => Seq(EagerSimpleFieldAccessInformationAnalysis)
-                    case Some(schedulers) => schedulers ++ Seq(EagerSimpleFieldAccessInformationAnalysis)
-                }
                 Seq(EagerSimpleFieldAccessInformationAnalysis)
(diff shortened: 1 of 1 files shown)


OPAL/br/src/main/scala/org/opalj/br/analyses/FieldAccessInformationKey.scala
```diff
@@ -26,6 +26,10 @@ object FieldAccessInformationKey extends ProjectInformationKey[FieldAccessInform
                     "analysis configuration",
                     s"no field access information analysis configured, using SimpleFieldAccessInformationAnalysis as a fallback"
                 )(project.logContext)
+                project.updateProjectInformationKeyInitializationData(this) {
+                    case None             => Seq(EagerSimpleFieldAccessInformationAnalysis)
+                    case Some(schedulers) => schedulers ++ Seq(EagerSimpleFieldAccessInformationAnalysis)
```
errt: This feels weird, since we'd still only return the EagerSimpleFieldAccessInformationAnalysis as the only requirement, so if there actually were other schedulers here, I suppose that could end up in a dead lock because of undeclared requirements.
Probably, this case should never happen anyway, because if initialization data was update while requirements are already computed, we could end up here with the compute method running different schedulers than the ones used here to compute the requirements.
Not sure what the best solution to this is. Keep yours? Replace this line with an exception? Replace the whole functionality here with getOrCreateProjectInformationKeyInitializationData?

maximilianruesch: I would either just use `getOrCreateProjectInformationKeyInitializationData` or throw an exception when a `Some(...)` is found, as the result handling of `getProjectInformationKeyInitializationData` and the update should happen in sync.

I will try to prevent this entirely by using `getOrCreateProjectInformationKeyInitializationData`.

maximilianruesch: It works, but we cannot give a compute function to this `getOrCreateProjectInformationKeyInitializationData` which means we lose the info log. Is that acceptable?

errt: Why? getOrCreateProjectInformationKeyInitializationData takes a by-name parameter, that should allow adding the logging?

maximilianruesch: I think I just had a brainfart, forgot that you can pass block expressions to everything in Scala 🤦

errt (commented):
LGTM

errt (approved):

PR #192, opened 2024-02-13
Change canonical form of single value integer range likes

maximilianruesch:
Provides better support for integer range likes with matching upper and lower bounds by converting them to a concrete `TheIntegerValue` as their canonical form.

> [!NOTE]
> We might want to implement this in more domain parts than the `IntegerRangeValues`, e.g. `IntegerSetValues` containing only one value. Any thoughts on this?

* Related to (and discovered during) #1


errt:
I agree that we should do something about this and we should make sure it is similar in all relevant domains. I'm not sure this is the right thing or place to do it, though. There's also `DefaultIntegerRangeValues` and its inner class IntegerRange which can model a full range, not just a single-element one. Then, there's also the `constantValue` method that is implemented in `IntegerSetValues.IntegerSetLike`, but not in `IntegerRangeValues.IntegerRangeLike`. I think we should look into these before settling on a solution.

errt (approved):
LGTM

PR #194, opened 2024-05-02
Add alias analysis

LukasKlenner:
Note: This pull request is based on #178 and should only be merged after that pull request.

This pull request adds two alias analyses for the alias properties added in #171: One intraprocedural alias analysis and one based on the existing points-to analysis.


OPAL/tac/src/main/scala/org/opalj/tac/fpcf/analyses/alias/AllocationSiteAndTacBasedAliasAnalysis.scala
```diff
@@ -0,0 +1,68 @@
...
+            val tac = state.tacai1.get
+
+            // the definition site is not the allocation site -> it is a method call or something similar
+            if (pc != defSite1.head) return false
+
+            for (stmt <- state.tacai1.get.stmts) {
+                stmt match {
+                    case goto: Goto =>
+                        val targetPC = tac.stmts(goto.targetStmt).pc
+                        if (targetPC <= pc && goto.pc >= pc) return false // jumping from behind the allocation site in front of it -> might break aliasing because allocation site is executed multiple times
```
errt: This is inefficient and incorrect:

Inefficient:
 - Checking every statement whether it is a goto, instead of using already available information (CFG, dominator tree)
 - Matching on the type of the instruction instead of its astID, which allows for an integer check instead of a type check
 - Extracting pcs instead of just working with stmt ids, which are ordered the same way

Incorrect:
 - Would interpret some linear control flow as loops (not unsound, but still):
```
...
5: goto 8
6: allocation site
7: goto 9
8: goto 6
...
```
 - Misses other branching instructions, such as if and switch, so it is unsound as it may miss loops

A better way to do this is to check whether a jump is dominated by its target. That is still not perfect (it misses non-natural loops). OPAL provides facilities to compute dominator trees and check dominance.

OPAL/tac/src/main/scala/org/opalj/tac/fpcf/analyses/alias/pointsto/TypePointsToBasedAliasAnalysisState.scala
```diff
@@ -0,0 +1,14 @@
+/* BSD 2-Clause License - see OPAL/LICENSE for details. */
```
errt: Please combine these small files into larger ones in a sensible way. E.g., this code should be in the file above.

OPAL/tac/src/main/scala/org/opalj/tac/fpcf/analyses/alias/pointsto/AbstractPointsToBasedAliasAnalysis.scala
```diff
@@ -0,0 +1,302 @@
...
+     * @param ase The [[AliasSourceElement]] the given pointsTo set is associated with.
+     * @param pointsToEntity The pointsTo entity the pointTo Set belongs to.
+     * @param pointsToSet The pointsTo set to handle.
+     */
+    private[this] def handlePointsToSet(ase: AliasSourceElement, pointsToEntity: Entity, pointsToSet: PointsToSet)(
+        implicit
+        state:   AnalysisState,
+        context: AnalysisContext
+    ): Unit = {
+        pointsToSet.forNewestNElements(pointsToSet.numElements - state.pointsToElementsHandled(ase, pointsToEntity)) {
```
errt: Why use a stored value of handled elements when you can just store the old points-to set and use its numElements?

OPAL/tac/src/main/scala/org/opalj/tac/fpcf/analyses/alias/package.scala
```diff
@@ -0,0 +1,41 @@
...
+    type AllocationSite = (Context, PC)
+
+    /**
+     * Converts the given [[UVar]] into an equivalent [[PersistentUVar]].
+     *
+     * @param uVar The UVar to convert.
+     * @param stmts The statements of the method the UVar is defined in.
+     * @return A persistent UVar for the given UVar.
+     */
+    final def persistentUVar(uVar: UVar[ValueInformation])(implicit stmts: Array[Stmt[V]]): PersistentUVar = {
```
errt: This doesn't look like it should be part of this package object (and neither does the rest of this file). Aren't they already available elsewhere? If not, place them in a more general place.

OPAL/tac/src/main/scala/org/opalj/tac/fpcf/analyses/alias/TypeBasedAliasSet.scala
```diff
@@ -0,0 +1,24 @@
+/* BSD 2-Clause License - see OPAL/LICENSE for details. */
+package org.opalj.tac.fpcf.analyses.alias
+
+import scala.collection.mutable
+
+import org.opalj.br.ReferenceType
+
+class TypeBasedAliasSet extends AliasSetLike[ReferenceType, TypeBasedAliasSet] {
+
+    private var _pointsTo: mutable.Set[ReferenceType] = mutable.Set.empty[ReferenceType]
```
errt: Why are you using mutable sets? You should really avoid mutability!

OPAL/tac/src/main/scala/org/opalj/tac/fpcf/analyses/alias/AliasSetLike.scala
```diff
@@ -0,0 +1,34 @@
...
+
+    def pointsToAny: Boolean = _pointsToAny
+
+    def isEmpty: Boolean = allPointsTo.isEmpty && !_pointsToAny
+
+    def size: Int = allPointsTo.size
+
+    def allPointsTo: mutable.Set[ElementType]
+
+    def intersection(other: T): T
```
errt: Why isn't this just implemented generically?

OPAL/tac/src/main/scala/org/opalj/tac/fpcf/analyses/alias/AllocationSiteBasedAliasSet.scala
```diff
@@ -0,0 +1,32 @@
...
+    }
+
+    def pointsTo(context: Context, pc: PC): Boolean = _pointsToAny || pointsTo((context, pc))
+
+    def allPointsTo: mutable.Set[AllocationSite] = _pointsTo
+
+    def intersection(other: AllocationSiteBasedAliasSet): AllocationSiteBasedAliasSet = {
+        val intersection = new AllocationSiteBasedAliasSet
+
+        // we could theoretically optimize this by stopping once two intersecting elements have been found
```
errt: Then why don't you do that? You could just change the method to return, e.g., `(Boolean, Boolean)` and replace the intersection with something like `_pointsTo.exists(other._pointsTo.contains)`.

OPAL/tac/src/main/scala/org/opalj/tac/fpcf/analyses/alias/AbstractAliasAnalysis.scala
```diff
@@ -0,0 +1,162 @@
...
+            return false
+
+        val type1 = getObjectType(context.element1.referenceType)
+        val type2 = getObjectType(context.element2.referenceType)
+        implicit val classHierarchy: ClassHierarchy = project.classHierarchy
+
+        if (type1.isEmpty || type2.isEmpty)
+            return false
+
+        if (type1.get._2 != type2.get._2)
```
errt: I think you should just be able to use ClassHierarchy.isSubtypeOf. It handles array types for you just well.

OPAL/br/src/main/scala/org/opalj/br/fpcf/properties/alias/AliasEntity.scala
```diff
@@ -0,0 +1,117 @@
...
+    private[this] val (_element1, _element2) = (e1, e2) match {
+        case (e1, e2) if e1.hashCode() < e2.hashCode() => (e1, e2)
+        case (e1, e2)                                  => (e2, e1)
+    }
+
+    /**
+     * A copy of two contexts of this [[AliasEntity]]. It uses the same order as the elements.
+     */
+    private[this] val (_context1, _context2) = (c1, c2) match {
+        case (c1, c2) if e1 == _element1 => (c1, c2)
```
errt: Use `eq`, not `==`. `==` invokes the `equals` method in Scala instead of performing a reference equality check

OPAL/tac/src/main/scala/org/opalj/tac/fpcf/analyses/alias/AllocationSiteAndTacBasedAliasAnalysis.scala
```diff
@@ -0,0 +1,68 @@
...
+
+        if (intersection.size != 1 ||
+            pointsTo1.size != 1 ||
+            pointsTo2.size != 1
+        ) return false
+
+        // they refer to the same allocation site but aren't necessarily the same object (e.g. if the allocation site
+        // is inside a loop or a different method and is executed multiple times)
+
+        val (pointsToContext: Context, pc: PC) = pointsTo1.allPointsTo.head
```
errt, line 29: You used `.head` here. This might be correct, but most often it isn't because it ignores the other elements.

OPAL/tac/src/main/scala/org/opalj/tac/fpcf/analyses/alias/AllocationSiteAndTacBasedAliasAnalysis.scala
LukasKlenner: If I understand it correct, a better way to do it would be iterate over all basic block with a endPC >= allocationSitePC (don't care what previous bb's do) and check whether they are dominated by a bb in front of the allocation site? I don't see a way to directly check "whether a jump is dominated by its target". That would still require me to iterate over all goto statements to identify them. Or should i keep iterating over the statements and only perform the check, whether the goto is problematic, using the dominatorTree? Regarding the implementation: To check the dominance I only found the DominatorTree.foreachDom(int)(int => U) method. How can I map the given int back to the associated bb to check whether that is a problematic dominance?

OPAL/tac/src/main/scala/org/opalj/tac/fpcf/analyses/alias/AllocationSiteAndTacBasedAliasAnalysis.scala
LukasKlenner: Before this line is executed we require that both pointsTo sets contain exactly one element. If that is not the case we directly return false. So that assumption shouldn't be a problem

errt: Is `pointsTo1.size` the same as `pointsTo1.allPointsTo.size`? Then yes

OPAL/tac/src/main/scala/org/opalj/tac/fpcf/analyses/alias/AllocationSiteAndTacBasedAliasAnalysis.scala
errt: You need to do basic blocks before the allocation, too. See the example above how (linear control flow) jumps can reorder code irrespective of program counters.
Jumps are only at the end of basic blocks, so you don't need to look for gotos (and also, you get all kinds of jumps not just gotos).
There is method `strictlyDominates` on DominatorTree (strictly just means it excludes the BB itself).
I can't come up with the correct procedure from my head just now, but I suppose it will be something like: Is there a dominator (loop header) of the allocation (loop body) which has a predecessor that it dominates (back edge).

OPAL/tac/src/main/scala/org/opalj/tac/fpcf/analyses/alias/AllocationSiteAndTacBasedAliasAnalysis.scala
LukasKlenner: Yes, it's the same

OPAL/tac/src/main/scala/org/opalj/tac/fpcf/analyses/alias/AllocationSiteAndTacBasedAliasAnalysis.scala
LukasKlenner: Sounds good. So far I have this code:

```
// check if the allocation site is dominated by a loop header, i.e., is executed multiple times
domTree.foreachDom(cfg.bb(tac.properStmtIndexForPC(pc)).nodeId)(dom => {

    // check if the dominator is a loop header
    cfg.foreachPredecessor(cfg.allBBs.toSeq(dom).nodeId)(pred => {

        // if the dominator itself dominates one of its predecessors, it is a loop header
        if (domTree.strictlyDominates(dom, cfg.bb(pred).nodeId)) {
            return false
        }
    })

}) 
```
It checks whether the allocation site is dominated by a loop header. It works so far, but overapproximates cases where the allocation site is after the loop and not inside of it. E.g.:
```
public static void mustAliasLoopInFront() {
    for (int i = 0; i < 10; i++) {
        [...]
    }

    Object o2 = new Object();

    o2.hashCode();
}
```
It computes that (o2,o2) is a may alias because the definition site is dominated by a loop header. Do you know whether it is possible to detect such cases, e.g., by checking if the allocation site is behind all predecessors of the loop header? Or would that cause other problems?

errt: I guess also check that the backedge is a postdominator of the allocation site?

LukasKlenner: I have tried to access the postDominator tree by adding the following method to the cfg class:

```
def postDominatorTree: PostDominatorTree = {
    val exitNodes = Seq.empty[Int] ++
        (if (normalReturnNode.predecessors.nonEmpty) Seq(normalReturnNode.nodeId) else Seq.empty) ++
        (if (abnormalReturnNode.predecessors.nonEmpty) Seq(abnormalReturnNode.nodeId) else Seq.empty)

    PostDominatorTree(
        if (exitNodes.size == 1) exitNodes.headOption else None,
        exitNodes.contains,
        IntTrieSet.empty,
        exitNodes.foreach,
        foreachSuccessor,
        foreachPredecessor,
        basicBlocks.last.endPC
    )
}
```
However, I get an exception when I try to use it:

```
Caused by: java.lang.ArrayIndexOutOfBoundsException: Index -1 out of bounds for length 10
	at org.opalj.graphs.DominatorTree$.$anonfun$create$1(DominatorTree.scala:192)
	at org.opalj.br.cfg.CFG.$anonfun$postDominatorTree$2(CFG.scala:575)
	at org.opalj.br.cfg.CFG.$anonfun$postDominatorTree$2$adapted(CFG.scala:575)
	at org.opalj.graphs.PostDominatorTree$.$anonfun$apply$3(PostDominatorTree.scala:155)
	at org.opalj.graphs.PostDominatorTree$.$anonfun$apply$3$adapted(PostDominatorTree.scala:155)
	at org.opalj.graphs.DominatorTree$.create(DominatorTree.scala:191)
	at org.opalj.graphs.PostDominatorTree$.apply(PostDominatorTree.scala:183)
	at org.opalj.br.cfg.CFG.postDominatorTree(CFG.scala:578)
        [...]
```

Do you know another way to access the PostDominatorTree or what the problem with this one could be?

errt: Not sure where that foreachSuccessor method you're using comes from, but apparently it returns -1 when it shouldn't. You can have a look at method postDominatorTree in org.opalj.ai.domain.RecordCFG to check how to properly compute post-dominator trees.

errt:
Please fix the merge conflicts listed below.

OPAL/tac/src/main/scala/org/opalj/tac/fpcf/analyses/alias/AllocationSiteAndTacBasedAliasAnalysis.scala
LukasKlenner: I have now found a way to access the postDominatorTree. It currently uses the same way used in the TACAIProvider class. But i am not sure if it is possible to reuse the results used for the tacai instead of calculating them again.

errt: I think that is okay for now. We can check whether that computation is performance-sensitive for the analysis, but I don't expect it to be as long as you don't recompute it multiple times per method.

OPAL/tac/src/main/scala/org/opalj/tac/fpcf/analyses/alias/AllocationSiteAndTacBasedAliasAnalysis.scala
```diff
@@ -41,21 +43,40 @@ trait AllocationSiteAndTacBasedAliasAnalysis extends AllocationSiteBasedAliasAna
             val defSite1 = context.element1.asAliasUVar.persistentUVar.defPCs
             val defSite2 = context.element2.asAliasUVar.persistentUVar.defPCs
 
-            if (defSite1.size != 1 || defSite1.size != 1 || defSite1.head != defSite2.head) return false // multiple or different def sites for one element -> might be different objects (e.g. due to recursion via parameter)
-
-            val tac = state.tacai1.get
+            // multiple or different def sites for one element -> might be different objects (e.g. due to recursion via parameter)
+            if (defSite1.size != 1 || defSite1.size != 1 || defSite1.head != defSite2.head) return false
```
errt: Should probably be defSite2 in the second condition.

OPAL/tac/src/main/scala/org/opalj/tac/fpcf/analyses/alias/AliasSetLike.scala
```diff
@@ -0,0 +1,111 @@
+/* BSD 2-Clause License - see OPAL/LICENSE for details. */
+package org.opalj.tac.fpcf.analyses.alias
```
errt: Fix package declarations please (org.opalj in one line, every other package on a line of its own), in several files in this PR

OPAL/tac/src/main/scala/org/opalj/tac/fpcf/analyses/alias/AllocationSiteAndTacBasedAliasAnalysis.scala
LukasKlenner: I have just implemented that they are being cached after the first computation

errt (changes requested):

OPAL/tac/src/main/scala/org/opalj/tac/fpcf/analyses/alias/AllocationSiteAndTacBasedAliasAnalysis.scala
errt: Note that TACode.cfg.dominatorTree is also recomputed every time and that doesn't seem to be a problem usually.

I think the code you have now will produce different kinds of trees for the dominator and post dominators: the dominator tree from the CFG should be a TAC-based one with Stmts in it, while the post dominator tree you compute is bytecode-based with bytecode Instructions in it, right? That might cause issues.

LukasKlenner: Good point, that's true (interesting that it still passes all tests* :D). I can of course map the values back to their pcs for the postdominator tree like so: `postDomTree.strictlyDominates(cfg.bb(dom).startPC, cfg.bb(allocBB).startPC)`. But a version of the PostDominatorTree that is based on the tac would probably be better.

\* They do fail if I change something else

errt:
I think I'm mostly happy with this PR now and will approve and merge it soon. Just have a final look into the post-dominator tree issue please. It seems very weird to do dominator and post dominator checks using different underlying representations. It would be much nicer to also use a TAC-based post-dominator tree.
I think your original strategy of borrowing from CFG's `dominatorTree` method should be the way to go, we'd just need to figure out why it failed in your first attempt. As far as I can tell, the `foreachSuccessor` method in CFG should never invoke the given function with -1, because the call in line 455 is guarded by the check in 453 and the one in 444 uses the next pc, which is the given pc + 1. It should probably be easy to debug where that -1 came from.
On another note, instead of caching the trees yourself, it might be worthwile to consider converting the `dominatorTree` (and `postDominatorTree`, once implemented) method on CFG to a `lazy val`. That way, no complex ad-hoc caching logic is needed and other analyses, etc., also using the dominator/post-dominator trees would also not need to recompute it.

LukasKlenner:
I have managed to fix my initial approach of adding a postDominatorTree method in the cfg class. The problem was that I directly used the id's of the special exit nodes, which are -1 and -2. I am now using the ids of their predecessors, which seems to work fine: `val exitNodes = normalReturnNode.predecessors.map(_.nodeId) ++ abnormalReturnNode.predecessors.map(_.nodeId)`. 

I have also converted the dominatorTree and postDominatorTree methods to `lazy val` as suggested.

However, the same test which made us use a postDominatorTree is still failing. The problem is, that an exception can be thrown inside the loop which creates an edge to the abnormal return node. This again causes the immediate post dominator of the loop header to be the unique exit node of the postDom tree instead of the bb after the loop, breaking our "behind-loop" check.

errt:
Can you have another look at this? It is good work and I'd like to merge it, but I can't as long as there are still failing tests.

LukasKlenner:
The remaining problem is the handling of abnormal returns inside the loop and I am not sure how to solve it. I would for now change the expected values of the two failing tests to MayAlias. If a more mature loop detection is implemented in the future this could be used here to improve the precision.

errt (commented):
Looks good now

errt (approved):

PR #201, opened 2024-07-26
Fixes toBinaryJavaName

errt:
Its semantics are given by https://docs.oracle.com/en/java/javase/22/docs/api/java.base/java/lang/Class.html#getName()

### DIFF
--- a/OPAL/br/src/main/scala/org/opalj/br/Type.scala
+++ b/OPAL/br/src/main/scala/org/opalj/br/Type.scala
@@ -188,10 +188,7 @@ sealed trait Type extends UIDValue with Ordered[Type] {
      * Returns the binary name of this type as used by the Java runtime. Basically
      * returns the same name as produced by `Class.getName`.
      */
-    @throws[UnsupportedOperationException](
-        "if this type has not binary name(i.e., if this type represents void)"
-    )
-    def toBinaryJavaName: String
+    def toBinaryJavaName: String = toJava
 
     /**
      * Returns the representation of this type as used by the JVM in, for example,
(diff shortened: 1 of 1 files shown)


OPAL/br/src/main/scala/org/opalj/br/Type.scala
```diff
@@ -555,7 +548,7 @@ sealed trait CTIntType extends Type {
  */
 case object CTIntType extends CTIntType {
     final val id = Int.MinValue + 1
-    def toBinaryJavaName: String = throw new UnsupportedOperationException()
+    override def toBinaryJavaName: String = throw new UnsupportedOperationException()
```
maximilianruesch, line 551: Why should int-like types not support this getter? They do, just like every other type (except array)
image

OPAL/br/src/main/scala/org/opalj/br/Type.scala
```diff
@@ -188,10 +188,7 @@ sealed trait Type extends UIDValue with Ordered[Type] {
      * Returns the binary name of this type as used by the Java runtime. Basically
      * returns the same name as produced by `Class.getName`.
```
maximilianruesch:
```suggestion
     * Returns the binary name of this type as used by the Java runtime.
     * Returns the same name as produced by `Class.getName`.
```
Lets update the comment so it says the intended truth.

maximilianruesch (changes requested):
Unsure about a small part.

Can we / should we add tests for this?

OPAL/br/src/main/scala/org/opalj/br/Type.scala
errt: They do? They inherit the default implementation toBinaryJavaName = toJava from `Type`. The line you marked is only the companion object that represents all computational int types and thus has no individual binary representation.

maximilianruesch: Damn didnt see that it was the companion object. Sometimes these confuse me when I do not have IDE support...

This is fine then.

maximilianruesch (approved):
deemed-excellent

PR #209, opened 2024-09-04
Remove absolute path to .scalafmt

errt:


maximilianruesch (approved):
Tested it on my local OPAL instance: Works

PR #210, opened 2024-09-04
Schedule the required FieldAccessInformationAnalysis in runners

errt:
We forgot to add them when the FieldAccessInformation became a necessary prerequisite for the FieldAssignability analyses

### DIFF
--- a/DEVELOPING_OPAL/tools/src/main/scala/org/opalj/support/info/Immutability.scala
+++ b/DEVELOPING_OPAL/tools/src/main/scala/org/opalj/support/info/Immutability.scala
@@ -65,9 +65,8 @@ import org.opalj.tac.cg.CallGraphKey
 import org.opalj.tac.cg.XTACallGraphKey
 import org.opalj.tac.fpcf.analyses.LazyFieldImmutabilityAnalysis
 import org.opalj.tac.fpcf.analyses.escape.LazySimpleEscapeAnalysis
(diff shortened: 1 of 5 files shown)


DEVELOPING_OPAL/tools/src/main/scala/org/opalj/support/info/Immutability.scala
```diff
@@ -157,8 +157,6 @@ object Immutability {
 
         project.get(callgraphKey)
 
-        L2PurityAnalysis.setRater(Some(SystemOutLoggingAllExceptionRater))
```
johannesduesing, line 160: I did not quite get how this relates to the PR - was this an intentional choice?

johannesduesing (commented):
Looks good to me, i ran tests (and the `FieldAssignabilityRunner` with a trivial example) on my machine and everything worked as expected 👍  I had just one minor question about a line that was removed.

DEVELOPING_OPAL/tools/src/main/scala/org/opalj/support/info/Immutability.scala
errt: Not really related to the PR, just happened to fix that as well. You could say it relates to it as it is a leftover from a previously required analysis and this PR corrects the setup of which analyses are really required now.

johannesduesing (approved):
Thanks for the explanation, i'm fine with this now 👍

PR #211, opened 2024-09-05
Add triggered field access information analysis and use for call graphs

maximilianruesch:
The current field access information analysis is eager. Using it in call graphs usually results in many upfront calculations, growing directly proportional to the number of methods under analysis, regardless of whether they are reachable or not.

This PR adds a triggered analysis for field accesses tor rectify this issue and thus brings the field access information analysis in line with other call graph related analysis and others that are used during call graph computation. Now, field accesses are only computed for methods that are deemed reachable, which is also the only case call graph analyses need such information.

Thus, the points to keys are reconfigured to use the new triggered analysis instead of the previous eager analysis.

### DIFF
--- a/OPAL/tac/src/main/scala/org/opalj/tac/fpcf/analyses/fieldaccess/FieldAccessInformationAnalysis.scala
+++ b/OPAL/tac/src/main/scala/org/opalj/tac/fpcf/analyses/fieldaccess/FieldAccessInformationAnalysis.scala
@@ -6,29 +6,38 @@ package analyses
 package fieldaccess
 
 import org.opalj.br.Method
+import org.opalj.br.analyses.DeclaredFields
 import org.opalj.br.analyses.DeclaredFieldsKey
+import org.opalj.br.analyses.DeclaredMethods
 import org.opalj.br.analyses.DeclaredMethodsKey
 import org.opalj.br.analyses.ProjectInformationKeys
 import org.opalj.br.analyses.SomeProject
 import org.opalj.br.fpcf.BasicFPCFEagerAnalysisScheduler
+import org.opalj.br.fpcf.BasicFPCFTriggeredAnalysisScheduler
 import org.opalj.br.fpcf.ContextProviderKey
 import org.opalj.br.fpcf.FPCFAnalysis
+import org.opalj.br.fpcf.FPCFAnalysisScheduler
+import org.opalj.br.fpcf.analyses.ContextProvider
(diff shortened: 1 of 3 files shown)


errt:
Looks good at a first glance, but please factor out common code between the two schedulers (e.g., the used and derived properties).

OPAL/tac/src/main/scala/org/opalj/tac/fpcf/analyses/fieldaccess/FieldAccessInformationAnalysis.scala
```diff
@@ -164,18 +202,47 @@ object EagerFieldAccessInformationAnalysis extends BasicFPCFEagerAnalysisSchedul
...
+    with BasicFPCFEagerAnalysisScheduler {
+
+    override def requiredProjectInformation: ProjectInformationKeys = super.requiredProjectInformation ++ Seq(
+        DeclaredMethodsKey,
+        ContextProviderKey
+    )
+
+    override def derivesEagerly: Set[PropertyBounds] = Set.empty
+
+    override def derivesCollaboratively: Set[PropertyBounds] = derivedProperties
```
errt, line 224: Why do you need to override these, can't you specify them in the super trait already? I wouldn't expect us to ever add a lazy scheduler where they wouldn't fit.

maximilianruesch: Since these functions stem from the specific `BasicEagerFPCFAnalysisScheduler` and `BasicTriggeredFPCFAnalysisScheduler` traits and not from the generic `FPCFAnalysisScheduler` trait that the super trait of these extends, they are correct here. It also only makes sense that you can specify e.g. eager behavior when you know you have an eager analysis.

errt: No, they stem from `ComputationSpecification`, the super trait of `FPCFAnalysisScheduler`. It's fine like this, though.

maximilianruesch: It seems the Github UI is not fit for determining the origin of super functions...

errt (approved):

PR #212, opened 2024-09-06
Turn assertion into logged error

errt:
In case of separately compiled class files, a method reference that should be in a class is actually found in an interface.

I didn't check if this would still be executable bytecode, but I think it is better for OPAL to continue trying to provide reasonable analysis results while warning users about the problem rather than throwing an assertion error in development mode and silently doing nothing in production mode.

### DIFF
--- a/OPAL/br/src/main/scala/org/opalj/br/analyses/ProjectLike.scala
+++ b/OPAL/br/src/main/scala/org/opalj/br/analyses/ProjectLike.scala
@@ -594,12 +594,12 @@ abstract class ProjectLike extends ClassFileRepository { project =>
 
         project.classFile(receiverType) match {
             case Some(classFile) =>
-                assert(
-                    !classFile.isInterfaceDeclaration, {
-                        val methodInfo = descriptor.toJava(receiverType.toJava, name)
-                        s"the method is defined in an interface $methodInfo"
-                    }
-                )
+                if (classFile.isInterfaceDeclaration) {
+                    OPALLogger.error(
+                        "project configuration - class hierarchy",
+                        s"invalid inheritance: $receiverType is an interface"
+                    )
+                }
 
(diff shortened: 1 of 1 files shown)


OPAL/br/src/main/scala/org/opalj/br/analyses/ProjectLike.scala
```diff
@@ -594,12 +594,12 @@ abstract class ProjectLike extends ClassFileRepository { project =>
 
         project.classFile(receiverType) match {
             case Some(classFile) =>
-                assert(
-                    !classFile.isInterfaceDeclaration, {
-                        val methodInfo = descriptor.toJava(receiverType.toJava, name)
-                        s"the method is defined in an interface $methodInfo"
-                    }
-                )
+                if (classFile.isInterfaceDeclaration) {
```
johannesduesing, line 597: Should we maybe update the JavaDoc comment (l.580, `[..] by a class (not interface) type.`) since interface types are now supported (although still not valid)?

johannesduesing (commented):
Looks good to me, just had one cosmetic question

OPAL/br/src/main/scala/org/opalj/br/analyses/ProjectLike.scala
errt: Valid point, but I think the Doc still correctly documents the method's purpose. Interfaces are not really supported now, we log the most severe type of error that OPAL has to really warn users that things are wrong here. We just don't throw an exeception based on the rather random condition of being a development or production build anymore.

johannesduesing: Fair enough, i'm on board with that 👍

johannesduesing (approved):
LGTM 👍

PR #213, opened 2024-09-07
Fix resolution of Points-To-CG modules

errt:
Actually, not all modules have `PointsTo` in their name (and the reference.conf was already set up to work like this).


maximilianruesch (approved):
Needed this fix locally, tried it and it works

PR #218, opened 2024-09-13
Make evaluationDepth thread-local and thus thread-safe

### LINKED ISSUE #208: Evaluation-Depth in `PKECPropertyStore` is not thread-safe
The `evaluationDepth` local field is used in the `doApply` method of the `PKECPropertyStore` and as such modified by all worker threads of this property store. Due to many threads usually being active at the same time, this variable might experience race conditions and consistency problems.

In my case, while working on #1, the `evaluationDepth` reached negative values up to `-300`, which completely defeated the purpose of the `MaxEvaluationDepth` limit that is configured in the `reference.conf`.

The aforementioned value should be made thread-safe by giving each thread their own value, since it is used to decide when to schedule computation tasks to a different thread. There might be no trivial way to go about this while maintaining cache coherence, but ignoring that one could just put up an array-like data structure containing values for all `PKECThreads` + the main thread (since forcing value computations outside of workers is done in e.g. the tests).

Since the `doApply` method is located outside the `PKECThreads`, a simple field inside this class cannot be used, and moving contexts around to the code inside `doApply` can be difficult. Obtaining a thread id (with the main thread obtaining the last one) can be done via:
```scala
val threadId = Thread.currentThread() match {
    case thread: PKECThread => threads.indexOf(thread)
    case _                  => THREAD_COUNT
}
```

> [!NOTE]
> Implementers of this issue might want to have a look at [ThreadLocal Variables](https://docs.oracle.com/javase/6/docs/api/java/lang/ThreadLocal.html), since wrapping the current variable in such an instance might just do the job.

errt:
Turns out, reading and writing a single variable across multiple threads without synchronization is actually a bad idea. And even if it worked, it would not be the correct semantics.

* Fixes #208

### DIFF
--- a/OPAL/si/src/main/scala/org/opalj/fpcf/par/PKECPropertyStore.scala
+++ b/OPAL/si/src/main/scala/org/opalj/fpcf/par/PKECPropertyStore.scala
@@ -38,7 +38,7 @@ class PKECPropertyStore(
 
     implicit val propertyStore: PKECPropertyStore = this
 
-    var evaluationDepth: Int = 0
+    val evaluationDepth: ThreadLocal[Int] = ThreadLocal.withInitial[Int](() => 0)
 
     val ps: Array[ConcurrentHashMap[Entity, EPKState]] =
         Array.fill(PropertyKind.SupportedPropertyKinds) { new ConcurrentHashMap() }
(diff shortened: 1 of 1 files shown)


johannesduesing (approved):
LGTM 👍

Issue #112, opened 2023-04-26
Triggered analyses not executed when running multiple computations


errt:
Triggered analyses are executed whenever the first property is computed for an entity/propertykey pair. If properties are pre-set in the property store, triggered analyses are also executed on the respective entities. However, triggered analyses are currently not executed when there are properties in the property store from a previous phase of analysis execution (i.e., when a fixed point has been reached and new analyses are started thereafter). This is inconsistent, surprising, and probably a bad idea in general.

PR #220, opened 2024-09-16
Improve interface of reachable method analysis and related states to prevent misuse

maximilianruesch:
During development on #1 a triggered analysis that is based on the `ReachableMethodAnalysis` trait got an eager scheduler. Turns out the `ReachableMethodAnalysis` trait implicitly expects to only implement triggered analyses and is not (yet) compatible with eager analyses.

It expects at least some `ub` on the `Callers` property on the entity (since that would be the only time it would be triggered), but with an eager analysis that might not be the case yet when the analysis values are computed at the same time as the call graph. The rest of the analysis actually returns a dependency on the callers property just fine, even if no callers are present: https://github.com/opalj/opal/blob/c6d8224f35629750092931809f6c0b9abb1ca3b3/OPAL/tac/src/main/scala/org/opalj/tac/fpcf/analyses/cg/ReachableMethodAnalysis.scala#L101

This PR updates the `ReachableMethodAnalysis` to stop throwing a match exception when used with an eager analysis and adds compatibility. At the same time, some other handling of the analysis is improved to add clarity or prevent further misuse (such as throwing around `null` values which are discouraged in Scala).

> [!NOTE]
> This PR does not yet update the usages of the reachable method analysis to not use the deprecated method since they can be pretty expansive. Actually, the whole deprecation is up for discussion, so a second PR changing all the usages may not be created yet until we agree on a solution (which can also be revert the changes made to the public interface).


errt:
Code changes look reasonable, but I'm not sure what the purpose of an eager ReachableMethodsAnalysis would be. ReachableMethodsAnalysis is meant for analyses that only want to analyze methods that are definitely reachable, and that is only known once there is a respective Callers property.

maximilianruesch:
That is true, now that I think about it. Lets look at it this way:

Initially, there is no purpose an eager `ReachableMethodAnalysis` could serve that a triggered one could not. But because of #112, using an eager analysis can be used as a workaround when computing properties outside of the computation of e.g. the call graph (like e.g. computing the `RTACallGraphKey` and then computing the string analysis from #1 in the next step).

Theoretically, we could just fix #112 and be fine. Since this code change does make sense regardless since it simply updates a class to reflect its own abilities, I'd still opt for merging this.

errt:
Agreed, but #112 should really be fixed, using eager schedulers to work around this is merely a band aid and will not be sufficient once analysis batching is implemented.

OPAL/tac/src/main/scala/org/opalj/tac/fpcf/analyses/cg/ReachableMethodAnalysis.scala
```diff
@@ -79,44 +68,54 @@ trait ReachableMethodAnalysis extends FPCFAnalysis with TypeConsumerAnalysis {
...
 
         Results(
-            InterimPartialResult(Set(eOptP), continuationForCallers(eOptP.ub, tacEP)),
+            InterimPartialResult(Set(eOptP), continuationForCallers(newCallers, tacEPOpt)),
             results
         )
     }
 
+    /**
+     * @deprecated Use [[processMethod]](ContextType, Option[_]) instead
```
errt: Doesn't this lead to compile errors since we have warnings-as-errors active? This should produce a deprecation warning after all for the existing subclasses, right?

maximilianruesch: That is true. I now decided to, instead of deprecating it, remove it entirely since we only do major updates for now anyways and the next release is allowed to break existing analyses. During this I will make it easier for analyses to define which portions of the reachable method analysis they implement (i.e. with / without body). This also includes a refactoring of the state handling for the only analysis that processes methods without a body but still implements `TACAIBasedAnalysisState` regardless  (which does not really make sense)...

OPAL/tac/src/main/scala/org/opalj/tac/fpcf/analyses/TACAIBasedAnalysisState.scala
```diff
@@ -26,35 +26,22 @@ trait TACAIBasedAnalysisState[TheContextType <: Context]
...
      * call to super!
      */
-    abstract override def dependees: Set[SomeEOptionP] = {
-        val otherDependees = super.dependees
-        if ((_tacDependee ne null) && _tacDependee.isRefinable)
-            otherDependees + _tacDependee
-        else
-            otherDependees
-    }
+    abstract override def dependees: Set[SomeEOptionP] = super.dependees ++ Some(_tacDependee).filter(_.isRefinable)
```
errt: I understand the wish to make this more 'functional', but I think this creates more overhead than necessary.

OPAL/tac/src/main/scala/org/opalj/tac/fpcf/analyses/cg/CGState.scala
```diff
@@ -38,3 +39,11 @@ class CGState[ContextType <: Context](
 
     def hasNonFinalCallSite: Boolean = _virtualCallSites.nonEmpty
 }
+
+/**
+ * @author Maximilian Rüsch
+ */
+class TACAIBasedCGState[ContextType <: Context](
+    override val callContext:                  ContextType,
```
errt: This looks weird. Does this really have to be an `override val`? Can't it just be a constructor parameter that is passed to the CGState constructor?

OPAL/tac/src/main/scala/org/opalj/tac/fpcf/analyses/cg/CallGraphAnalysis.scala
```diff
@@ -121,25 +120,27 @@ class CallGraphAnalysis private[cg] (
...
-        val state = new CGState[ContextType](callContext, tacEP)
-        if (tacEP ne null)
+        if (tacEPOpt.isDefined) {
+            val state = new TACAIBasedCGState[ContextType](callContext, tacEPOpt.get)
             processMethod(state, new DirectCalls())
-        else
-            returnResult(new DirectCalls(), enforceCalleesResult = true)(state)
+        } else {
+            val state = new CGState[ContextType](callContext)
+            Results {
```
errt: Use round parentheses

OPAL/tac/src/main/scala/org/opalj/tac/fpcf/analyses/cg/CallGraphAnalysis.scala
```diff
@@ -121,25 +120,27 @@ class CallGraphAnalysis private[cg] (
...
+                new DirectCalls().partialResults(state.callContext, enforceCalleesResult = true)
```
errt: Does this actually do anything? Doesn't look like it.
I think this produces no results at all while missing what the original `returnResults` call would have returned: the InterimPartialResult for the state's dependees.

OPAL/tac/src/main/scala/org/opalj/tac/fpcf/analyses/cg/CallGraphAnalysis.scala
```diff
@@ -121,25 +120,27 @@ class CallGraphAnalysis private[cg] (
...
     ): ProperPropertyComputationResult = {
-        val state = new CGState[ContextType](callContext, tacEP)
-        if (tacEP ne null)
+        if (tacEPOpt.isDefined) {
+            val state = new TACAIBasedCGState[ContextType](callContext, tacEPOpt.get)
             processMethod(state, new DirectCalls())
-        else
-            returnResult(new DirectCalls(), enforceCalleesResult = true)(state)
+        } else {
+            val state = new CGState[ContextType](callContext)
```
errt: Having two different kinds of state here looks strange. If the code below actually did anything (see other comment), I'd expect this to fail in the continuation, because the continuation would get an unexpected kind of state.

OPAL/tac/src/main/scala/org/opalj/tac/fpcf/analyses/cg/ReachableMethodAnalysis.scala
```diff
@@ -36,18 +35,9 @@ trait ReachableMethodAnalysis extends FPCFAnalysis with TypeConsumerAnalysis {
...
-            case eps: EPS[_, _] =>
-                if (eps.ub eq NoCallers) {
-                    // we can not create a dependency here, so the analysis is not allowed to create
-                    // such a result
-                    throw new IllegalStateException("illegal immediate result for callers")
-                }
-            // the method is reachable, so we analyze it!
+
+        if (callersEOptP.isFinal && callersEOptP.ub == NoCallers) {
+            return NoResult;
```
errt, line 42: Shouldn't we keep the check for an illegal `NoCallers` interim state at least as an assertion? Right now it is not clearly visible that the method is reachable and should be analyzed.

OPAL/tac/src/main/scala/org/opalj/tac/fpcf/analyses/cg/ReachableMethodAnalysis.scala
```diff
@@ -65,47 +55,41 @@ trait ReachableMethodAnalysis extends FPCFAnalysis with TypeConsumerAnalysis {
...
 
     private[this] def processMethod(
-        eOptP: EOptionP[DeclaredMethod, Callers],
-        seen:  Callers,
-        tacEP: EPS[Method, TACAI]
+        eOptP:      EOptionP[DeclaredMethod, Callers],
+        oldCallers: Callers,
+        tacEPOpt:   Option[EPS[Method, TACAI]]
     ): ProperPropertyComputationResult = {
+        val newCallers = if (eOptP.hasUBP) eOptP.ub else NoCallers
```
errt, line 80: Wouldn't it be more sensible to turn this into an if, only executing the `forNewCalleeContexts` if there is a UBP?

OPAL/tac/src/main/scala/org/opalj/tac/fpcf/analyses/cg/ReachableMethodAnalysis.scala
```diff
@@ -65,47 +55,41 @@ trait ReachableMethodAnalysis extends FPCFAnalysis with TypeConsumerAnalysis {
...
-    }
+    ): PropertyComputationResult = processMethod(eOptP, NoCallers, None)
 
     private[this] def processMethod(
-        eOptP: EOptionP[DeclaredMethod, Callers],
-        seen:  Callers,
-        tacEP: EPS[Method, TACAI]
+        eOptP:      EOptionP[DeclaredMethod, Callers],
+        oldCallers: Callers,
+        tacEPOpt:   Option[EPS[Method, TACAI]]
```
errt: An EPS wrapped into an Option looks wrong. EPS is already a subtype of EOptionP which captures the possibility of no value being present.

OPAL/tac/src/main/scala/org/opalj/tac/fpcf/analyses/cg/ReachableMethodAnalysis.scala
```diff
@@ -125,12 +109,28 @@ trait ReachableMethodAnalysis extends FPCFAnalysis with TypeConsumerAnalysis {
...
     )(
         update: SomeEPS
     ): ProperPropertyComputationResult = {
         val newCallers = update.asInstanceOf[EPS[DeclaredMethod, Callers]]
-        processMethod(newCallers, oldCallers, tacEP)
+        processMethod(newCallers, oldCallers, tacEPOpt)
     }
+}
 
+trait DefinedBodyReachableMethodAnalysis extends ReachableMethodAnalysis {
```
errt: Since there seems to be only one ReachableMethodAnalysis that does process methods without body, might it be more sensible to switch the logic around, with the base trait not doing that and the specialized trait doing it?

errt (changes requested):

OPAL/tac/src/main/scala/org/opalj/tac/fpcf/analyses/cg/ReachableMethodAnalysis.scala
maximilianruesch: The `NoCallers` interim state was illegal since the code at the top did not know the class could use the continuation without a depending caller. Since the class can invoke and handle continuations just fine without using a depender constructed from a caller, the `NoCallers` interim state is not illegal anymore.

OPAL/tac/src/main/scala/org/opalj/tac/fpcf/analyses/cg/ReachableMethodAnalysis.scala
maximilianruesch: Since we need the `newCallers` for constructing a continuation and `forNewCalleeContexts`  does not do anything when `NoCallers` is given, I view this as reasonable and less verbose as well.

OPAL/tac/src/main/scala/org/opalj/tac/fpcf/analyses/cg/ReachableMethodAnalysis.scala
errt: But it's an illegal state in general, isn't it? I don't think call graph analyses are supposed to ever create such an interim state.

OPAL/tac/src/main/scala/org/opalj/tac/fpcf/analyses/cg/ReachableMethodAnalysis.scala
maximilianruesch: It kind of depends on whether you understand `EPK` as an interim / refinable state (which it says it is since the class expresses it is not final) that *may*  produce a value if awaited via a continuation OR you understand it as a simple "there is no property here and there will never be one".

Since my understanding of the class is the former, I think an `Option[EPS]` is more reasonable than an `EOptionP` here.

OPAL/tac/src/main/scala/org/opalj/tac/fpcf/analyses/cg/ReachableMethodAnalysis.scala
maximilianruesch: If it is an illegal state in general, then it should be asserted by the call graph generating framework and not the framework for analysis that use the call graph (even if they at least partially overlap). This class seems like the wrong area of code to make such an assertion (e.g. it would only be triggered when using such an analysis even though the result should be illegal regardless of using this class or not).

OPAL/tac/src/main/scala/org/opalj/tac/fpcf/analyses/cg/ReachableMethodAnalysis.scala
errt: Well, `EPK` is also used for queries, just representing a pair of entity and propertykey as the name suggest.

OPAL/tac/src/main/scala/org/opalj/tac/fpcf/analyses/cg/CallGraphAnalysis.scala
maximilianruesch: It does do something: Enforce a partial callees result to be produced as it would by an empty direct calls helper instance. If you look at the previous code, it was simply calling `returnResult` with the state and a newly instantiated helper object. Since the state cannot have a dependee since we did not provide it any (and it does not have access to the project or propertyStore), no `InterimResult` could be ever produced. Thus, we can specialize `returnResult` to use the TACAI-Based state to use in the continuation and return our result independently here.

I will make this clearer by removing the newly created state entirely, since we only use the `callContext` from it anyways.

OPAL/tac/src/main/scala/org/opalj/tac/fpcf/analyses/cg/CallGraphAnalysis.scala
maximilianruesch: Since there is no continuation (see https://github.com/opalj/opal/pull/220#discussion_r1766541607), it cannot fail. I will remove the state entirely (see the other comment again).

OPAL/tac/src/main/scala/org/opalj/tac/fpcf/analyses/cg/ReachableMethodAnalysis.scala
maximilianruesch: I pondered this and did could not answer this question. On one hand, if the specialized trait is used far more than the base trait, something might be wrong. Additionally, the specialized trait right now restricts the abilities of the base trait (if you want to view "does not process but just returns no result" as a restriction) which feels a bit against the core principles of inheritance.

On the other hand, even if the base trait does not handle unknown bodies, it still needs to define extension points for
1. The unknown body specialised trait to handle such unknown bodies
2. A `processMethod` or similar that has an `EOptionP` as a parameter.

In short (using the flipped version): If you only extend the `ReachableMethodAnalysis` you should only need to implement the `EPS` version of `processMethod` since you know you have a body. If you extend the `UnknownBodyReachableMethodAnalysis`, you should only need to implement the `EOptionP` version of `processMethod` which can handle both scenarios.

All in all, I feel more comfortable with the current solution, but I am happy to replace it with a better one if you have some ideas.

OPAL/tac/src/main/scala/org/opalj/tac/fpcf/analyses/cg/CallGraphAnalysis.scala
errt: Oh, right, I didn't see that.

OPAL/tac/src/main/scala/org/opalj/tac/fpcf/analyses/cg/ReachableMethodAnalysis.scala
maximilianruesch: I tried to use an `EPK`, but there is a problem: Since the `processMethodWithoutBody` is also invoked when no single defined method exists, there may be no `Method` to derive from the given `DeclaredMethod` with which you would create an `EPK[Method, TACAI]`. See https://github.com/opalj/opal/actions/runs/10939312309/job/30369296437?pr=220.

So we probably have to go back to using an `Option`, be it `Option[EPS]` or `Option[EOptionP]`. Or do you have another idea?

errt:
I went over this PR and tried several options, none satisfactory. I ended up moving the `null` based handling of methods without bodies into the only analysis that actually uses it. This avoids having `Option[EOptionP[...]]` or `EPK`s that will never be computed. It still uses `null`, but not as part of the interface of `ReachableMethodAnalysis`, but only internally in `CallGraphAnalysis`. This seems the cleanest solution for me for now, but please voice your opinions on this.

maximilianruesch:
I think moving the null handling into the `CallGraphAnalysis` is a good idea. In the current version however, the null value still traverses into code from the `ReachableMethodsAnalysis` trait and might cause unexpected behaviour. Lets take your idea one step further.

Surprisingly, with slightly reshuffling processing of methods, the API that needs adaptation for new analyses to support methods without bodies is now as close to the "with-body" analogy as can be: The same interface, simply without a TAC EP.
Additionally, this has the benefit of removing the null value invocation alltogether. I have attached a sample patch (inline, since uploading `.patch` files from Linux is currently not supported on GitHub), so please have a look. If youd like me to push it, ill be happy to.

```patch
Subject: [PATCH] Reduce null escape surface in reachable method analysis
---
Index: OPAL/tac/src/main/scala/org/opalj/tac/fpcf/analyses/cg/CallGraphAnalysis.scala
<+>UTF-8
===================================================================
diff --git a/OPAL/tac/src/main/scala/org/opalj/tac/fpcf/analyses/cg/CallGraphAnalysis.scala b/OPAL/tac/src/main/scala/org/opalj/tac/fpcf/analyses/cg/CallGraphAnalysis.scala
--- a/OPAL/tac/src/main/scala/org/opalj/tac/fpcf/analyses/cg/CallGraphAnalysis.scala	(revision 2c2043406aa0cd849b07a9a46c2329adeeff7bcd)
+++ b/OPAL/tac/src/main/scala/org/opalj/tac/fpcf/analyses/cg/CallGraphAnalysis.scala	(revision edaf1d76c78dc223a4e2d06cdc8f4c95eb2d3c19)
@@ -5,7 +5,6 @@
 package analyses
 package cg
 
-import org.opalj.br.DeclaredMethod
 import org.opalj.br.Method
 import org.opalj.br.MethodDescriptor
 import org.opalj.br.ObjectType
@@ -21,7 +20,6 @@
 import org.opalj.br.fpcf.properties.cg.Callers
 import org.opalj.br.fpcf.properties.cg.OnlyCallersWithUnknownContext
 import org.opalj.fpcf.Entity
-import org.opalj.fpcf.EOptionP
 import org.opalj.fpcf.EPK
 import org.opalj.fpcf.EPS
 import org.opalj.fpcf.InterimEUBP
@@ -29,7 +27,6 @@
 import org.opalj.fpcf.InterimUBP
 import org.opalj.fpcf.ProperPropertyComputationResult
 import org.opalj.fpcf.PropertyBounds
-import org.opalj.fpcf.PropertyComputationResult
 import org.opalj.fpcf.PropertyKind
 import org.opalj.fpcf.PropertyStore
 import org.opalj.fpcf.Results
@@ -121,20 +
IDEA additional info:
Subsystem: com.intellij.openapi.diff.impl.patch.CharsetEP118,16 @@
         }
     }
 
-    override final def processMethodWithoutBody(eOptP: EOptionP[DeclaredMethod, Callers]): PropertyComputationResult = {
-        processMethodWithoutBody(eOptP, null)
+    override final def processMethodWithoutBody(callContext: ContextType): ProperPropertyComputationResult = {
+        Results(new DirectCalls().partialResults(callContext, enforceCalleesResult = true))
     }
 
     override final def processMethod(
         callContext: ContextType,
         tacEP:       EPS[Method, TACAI]
     ): ProperPropertyComputationResult = {
-        if (tacEP ne null) {
-            val state = new TACAIBasedCGState[ContextType](callContext, tacEP)
-            processMethod(state, new DirectCalls())
-        } else {
-            Results(new DirectCalls().partialResults(callContext, enforceCalleesResult = true))
-        }
+        val state = new TACAIBasedCGState[ContextType](callContext, tacEP)
+        processMethod(state, new DirectCalls())
     }
 
     protected[this] def doHandleVirtualCall(
Index: OPAL/tac/src/main/scala/org/opalj/tac/fpcf/analyses/cg/ReachableMethodAnalysis.scala
<+>UTF-8
===================================================================
diff --git a/OPAL/tac/src/main/scala/org/opalj/tac/fpcf/analyses/cg/ReachableMethodAnalysis.scala b/OPAL/tac/src/main/scala/org/opalj/tac/fpcf/analyses/cg/ReachableMethodAnalysis.scala
--- a/OPAL/tac/src/main/scala/org/opalj/tac/fpcf/analyses/cg/ReachableMethodAnalysis.scala	(revision 2c2043406aa0cd849b07a9a46c2329adeeff7bcd)
+++ b/OPAL/tac/src/main/scala/org/opalj/tac/fpcf/analyses/cg/ReachableMethodAnalysis.scala	(revision edaf1d76c78dc223a4e2d06cdc8f4c95eb2d3c19)
@@ -27,7 +27,7 @@
  * Base trait for analyses that are executed for every method that is reachable.
  * The analysis is performed by `processMethod`.
  *
- * Note that methods without a body are not processed unless `processMethodWithoutBody` is overridden
+ * Note that methods without a body are not processed unless `processMethodWithoutBody` is overridden.
  *
  * @author Florian Kuebler
  */
@@ -44,7 +44,7 @@
 
         // we only allow defined methods
         if (!declaredMethod.hasSingleDefinedMethod)
-            return processMethodWithoutBody(callersEOptP);
+            return processNewContexts(callersEOptP, NoCallers) { processMethodWithoutBody };
 
         val method = declaredMethod.definedMethod
 
@@ -52,34 +52,31 @@
         if (method.classFile.thisType != declaredMethod.declaringClassType)
             return NoResult;
 
-        if (method.body.isEmpty)
+        if (method.body.isEmpty) {
             // happens in particular for native methods
-            return processMethodWithoutBody(callersEOptP);
+            return processNewContexts(callersEOptP, NoCallers) { processMethodWithoutBody };
+        };
 
         val tacEP = propertyStore(method, TACAI.key)
         if (tacEP.hasUBP && tacEP.ub.tac.isDefined) {
-            processMethod(callersEOptP, NoCallers, tacEP.asEPS)
+            processMethodWithTAC(callersEOptP, NoCallers, tacEP.asEPS)
         } else {
             InterimPartialResult(Set(tacEP), continuationForTAC(declaredMethod))
         }
     }
 
-    protected def processMethodWithoutBody(
-        eOptP: EOptionP[DeclaredMethod, Callers]
-    ): PropertyComputationResult = NoResult
-
-    protected def processMethodWithoutBody(
-        eOptP: EOptionP[DeclaredMethod, Callers],
-        tacEP: EPS[Method, TACAI]
+    private def processMethodWithTAC(
+        eOptP:      EOptionP[DeclaredMethod, Callers],
+        oldCallers: Callers,
+        tacEP:      EPS[Method, TACAI]
     ): ProperPropertyComputationResult = {
-        processMethod(eOptP, NoCallers, tacEP)
+        processNewContexts(eOptP, oldCallers) { processMethod(_, tacEP) }
     }
 
-    protected def processMethod(
+    private def processNewContexts(
         eOptP:      EOptionP[DeclaredMethod, Callers],
         oldCallers: Callers,
-        tacEP:      EPS[Method, TACAI]
-    ): ProperPropertyComputationResult = {
+    )(processMethod: ContextType => PropertyComputationResult): ProperPropertyComputationResult = {
         val newCallers = if (eOptP.hasUBP) eOptP.ub else NoCallers
         var results: List[ProperPropertyComputationResult] = Nil
 
@@ -87,15 +84,19 @@
             val theCalleeContext =
                 if (calleeContext.hasContext) calleeContext.asInstanceOf[ContextType]
                 else typeIterator.newContext(eOptP.e)
-            results ::= processMethod(theCalleeContext, tacEP)
+            results ::= processMethod(theCalleeContext)
         }
 
         Results(
-            InterimPartialResult(Set(eOptP), continuationForCallers(newCallers, tacEP)),
+            InterimPartialResult(Set(eOptP), continuationForCallers(processNewContexts(_, newCallers)(processMethod))),
             results
         )
     }
 
+    protected def processMethodWithoutBody(
+        callContext: ContextType,
+    ): PropertyComputationResult = NoResult
+
     def processMethod(
         callContext: ContextType,
         tacEP:       EPS[Method, TACAI]
@@ -106,7 +107,7 @@
     )(someEPS: SomeEPS): ProperPropertyComputationResult = {
         someEPS match {
             case UBP(tac: TACAI) if tac.tac.isDefined =>
-                processMethod(
+                processMethodWithTAC(
                     propertyStore(declaredMethod, Callers.key),
                     NoCallers,
                     someEPS.asInstanceOf[EPS[Method, TACAI]]
@@ -117,12 +118,8 @@
     }
 
     private[this] def continuationForCallers(
-        oldCallers: Callers,
-        tacEP:      EPS[Method, TACAI]
-    )(
-        update: SomeEPS
-    ): ProperPropertyComputationResult = {
-        val newCallers = update.asInstanceOf[EPS[DeclaredMethod, Callers]]
-        processMethod(newCallers, oldCallers, tacEP)
+        processMethod: EOptionP[DeclaredMethod, Callers] => ProperPropertyComputationResult
+    )(someEPS: SomeEPS): ProperPropertyComputationResult = {
+        processMethod(someEPS.asInstanceOf[EPS[DeclaredMethod, Callers]])
     }
 }
```

errt:
@maximilianruesch Thank you, this looks sensible. I was looking for a way to reuse the method that you now named `processNewContexts`, but didn't find a suitable way. You're right that it works nicely with the closure. You can push the change if you'd like to.

maximilianruesch:
I had to adjust the return type of the `processMethodWithoutBody` slightly to be able to properly assert when it does not return a real result (since `Results` only accepts lists of actual results, i.e. `NoResult` is not allowed).

As an alternative to the current solution with `Option[ProperPropertyComputationResult]` we could revert to `PropertyComputationResult` and instead introduce  `isProperResult` (or similar) on the latter to allow for testing if an object is `NoResult` (without an instance check). Let me know what you think.

Otherwise, this PR should be good to go!

errt:
It would probably be simpler to to just return something that is a `ProperPropertyComputationResult` but does nothing. E.g. an empty `Results` or `InterimPartialResult`.

errt (dismissed):
LGTM now

errt (approved):

PR #226, opened 2024-10-26
Improving Command-Parsing in OPAL with Scallop-Library

thanhtung24:


DEVELOPING_OPAL/tools/src/main/scala/org/opalj/support/info/CallGraph.scala
```diff
@@ -56,6 +56,10 @@ import org.opalj.tac.fpcf.analyses.pointsto.TamiFlexKey
 import org.opalj.util.PerformanceEvaluation.time
 import org.opalj.util.Seconds
 
+class CallGraphConf(args: Array[String]) {
```
errt: What is this supposed to be?

DEVELOPING_OPAL/tools/src/main/scala/org/opalj/support/parser/AnalysisCommandExternalParser.scala
```diff
@@ -0,0 +1,23 @@
...
+import org.opalj.Commandline_base.commandlines.OpalCommandExternalParser
+import org.opalj.br.fpcf.FPCFLazyAnalysisScheduler
+import org.opalj.br.fpcf.analyses.LazyL0PurityAnalysis
+import org.opalj.tac.fpcf.analyses.purity.{LazyL1PurityAnalysis, LazyL2PurityAnalysis}
+
+/**
+ * `AnalysisCommandExternalParser` interprets a command-line argument specifying a purity analysis level
+ * and returns the corresponding `FPCFLazyAnalysisScheduler`.
+ */
+object AnalysisCommandExternalParser extends OpalCommandExternalParser[FPCFLazyAnalysisScheduler]{
```
errt: Confusing naming. If it is only for purity, the name should say so. Also, in that case maybe make it an inner class in the purity runner.

DEVELOPING_OPAL/tools/src/main/scala/org/opalj/support/parser/AnalysisCommandExternalParser.scala
```diff
@@ -0,0 +1,23 @@
...
+    override def parse[T](arg: T): FPCFLazyAnalysisScheduler = {
+        val analysisName = arg.asInstanceOf[String]
```
errt: This isn't a good idea. Make the T a type parameter of OpalCommandExternalParser, then make the class here extend OpalCommandExternalParser[String, FPCFLazyAnalysisScheduler].
Or - can there ever be a case where the type here is not String? I couldn't find one in this PR at least. So why have the genericity at all?

DEVELOPING_OPAL/tools/src/main/scala/org/opalj/support/parser/AnalysisCommandExternalParser.scala
```diff
@@ -0,0 +1,23 @@
+package org.opalj.support.parser
+
+import org.opalj.Commandline_base.commandlines.OpalCommandExternalParser
+import org.opalj.br.fpcf.FPCFLazyAnalysisScheduler
+import org.opalj.br.fpcf.analyses.LazyL0PurityAnalysis
+import org.opalj.tac.fpcf.analyses.purity.{LazyL1PurityAnalysis, LazyL2PurityAnalysis}
```
errt: Do not use multi imports!

DEVELOPING_OPAL/tools/src/main/scala/org/opalj/support/parser/AnalysisCommandExternalParser.scala
```diff
@@ -0,0 +1,23 @@
+package org.opalj.support.parser
```
errt: Break your package declarations into multiple lines.
First line always is `package org.opalj`, then each subpackage gets a line of its own, e.g. `package support`

errt: Add licence header to your files

DEVELOPING_OPAL/tools/src/main/scala/org/opalj/support/parser/ClassPathCommandExternalParser.scala
```diff
@@ -0,0 +1,61 @@
...
+ * `ClassPathCommandExternalParser` parses and validates a class path string.
+ * It processes the class path provided as a command-line argument, splitting it by the system path
+ * separator (":" on UNIX, ";" on Windows), and verifies each entry, returning a sequence of valid
+ * `File` objects.
+ */
+object ClassPathCommandExternalParser extends OpalCommandExternalParser[Seq[File]]{
+    override def parse[T](arg: T): Seq[File] = {
+        val classPath = arg.asInstanceOf[String]
+        var cp = IndexedSeq.empty[String]
+        import scala.collection.immutable.ArraySeq
```
errt: Put imports at the top of the file

DEVELOPING_OPAL/tools/src/main/scala/org/opalj/support/parser/ClassPathCommandExternalParser.scala
```diff
@@ -0,0 +1,61 @@
...
+
+        cp = ArraySeq.unsafeWrapArray(
+            classPath.substring(classPath.indexOf('=') + 1).split(File.pathSeparator)
+        )
+
+        if (cp.isEmpty) cp = ArraySeq.unsafeWrapArray(Array(System.getProperty("user.dir")))
```
errt: Formatting: line break after if condition

errt: I don't think you need unsafeWrapArray here, because you can directly create a Seq that contains the value.

DEVELOPING_OPAL/tools/src/main/scala/org/opalj/support/parser/ClassPathCommandExternalParser.scala
```diff
@@ -0,0 +1,61 @@
...
+ * `File` objects.
+ */
+object ClassPathCommandExternalParser extends OpalCommandExternalParser[Seq[File]]{
+    override def parse[T](arg: T): Seq[File] = {
+        val classPath = arg.asInstanceOf[String]
+        var cp = IndexedSeq.empty[String]
+        import scala.collection.immutable.ArraySeq
+
+        cp = ArraySeq.unsafeWrapArray(
+            classPath.substring(classPath.indexOf('=') + 1).split(File.pathSeparator)
```
errt: Could probably just to `.split(...).toSeq` here instead of the unsafeWrapArray

OPAL/br/src/main/scala/org/opalj/br/analyses/AnalysisApplication.scala
```diff
@@ -134,7 +134,7 @@ trait AnalysisApplication {
             } else if (arg == "-renderConfig") {
                 renderConfig = true
             } else {
-                unknownArgs ::= arg
+                    unknownArgs ::= arg
```
errt: Broken formatting. Run `sbt scalafmt` on your code please.

OPAL/common/src/main/scala/org/opalj/Commandline_base/commandlines/AnalysisCommand.scala
```diff
@@ -0,0 +1,12 @@
+package org.opalj.Commandline_base.commandlines
```
errt: Fix package name. Package names should not start with capital letters nor should they contain underscores.

OPAL/project/build.properties
```diff
@@ -0,0 +1 @@
+sbt.version=1.9.7
```
errt: This file obviously was moved/copied to a wrong place. Remove it

project/Dependencies.scala
```diff
@@ -71,15 +73,14 @@ object Dependencies {
...
 
-  def common(scalaVersion: String) = Seq(reflect(scalaVersion), scalaparallelcollections, scalaxml, playjson, ficus, fastutil)
+  def common(scalaVersion: String) = Seq(reflect(scalaVersion), scalaparallelcollections, scalaxml, playjson, ficus, fastutil, scallop)
 
   val si = Seq()
   val bi = Seq(commonstext)
   val br = Seq(scalaparsercombinators, scalaxml)
   val ifds = Seq()
-  val tools = Seq(txtmark, jacksonDF)
+  val tools = Seq(txtmark, jacksonDF, scallop)
```
errt: I think if common has that dependency, you don't need it again in tools?

OPAL/common/src/main/scala/org/opalj/Commandline_base/commandlines/OpalConf.scala
```diff
@@ -0,0 +1,33 @@
...
+import org.rogach.scallop.exceptions.{Help, ScallopException}
+import org.rogach.scallop.{ScallopConf, ScallopOption, ValueConverter}
+
+/**
+ * `OpalConf` is a utility trait designed to enhance the `ScallopConf` functionality by providing additional methods
+ * for handling command-line argument parsing in the OPAL framework. This trait is intended to be mixed into
+ * `ScallopConf` configurations, enabling more customized parsing mechanisms with reusable methods for specific command
+ * types and error handling.
+ */
+trait OpalConf {self: ScallopConf =>
```
errt: Formatting: Line break after opening brace

OPAL/common/src/main/scala/org/opalj/Commandline_base/commandlines/OpalConf.scala
```diff
@@ -0,0 +1,33 @@
...
+    def parseCommand[T](command: ScallopOption[T]): Option[T] = if (!command.isDefined) None else Some(command.apply())
+
+    def parseCommandWithInternalParser[T, R](command: ScallopOption[T], internalParser: OpalCommand): Option[R] = if (!command.isDefined) None else Some(internalParser.parse(command.apply()).asInstanceOf[R])
+
+    def parseCommandWithExternalParser[T, R](command: ScallopOption[T], externalParser: OpalCommandExternalParser[R]): Option[R] = if (!command.isDefined) None else Some(externalParser.parse(command.apply()).asInstanceOf[R])
+
+    override def onError(e: Throwable): Unit = e match {
+        case Help("") => printHelp()
+        case ScallopException(message) =>
+            println(s"Error: $message")
```
errt: Please use the OPALLogger facilities instead of println

DEVELOPING_OPAL/tools/src/main/scala/org/opalj/support/info/Purity.scala
```diff
@@ -77,9 +73,139 @@ import org.opalj.tac.fpcf.analyses.purity.L1PurityAnalysis
 import org.opalj.tac.fpcf.analyses.purity.L2PurityAnalysis
 import org.opalj.tac.fpcf.analyses.purity.LazyL1PurityAnalysis
 import org.opalj.tac.fpcf.analyses.purity.LazyL2PurityAnalysis
-import org.opalj.tac.fpcf.analyses.purity.SystemOutLoggingAllExceptionRater
 import org.opalj.util.PerformanceEvaluation.time
 import org.opalj.util.Seconds
+import org.rogach.scallop.ScallopConf
+
+/**
+ * `PurityConf` is a configuration class for parsing and managing command-line arguments related to purity analysis
```
errt: There's a merge conflict on this file, please merge the current state of develop into your branch and fix the conflict.

DEVELOPING_OPAL/tools/src/main/scala/org/opalj/support/info/Purity.scala
```diff
@@ -77,9 +73,139 @@ import org.opalj.tac.fpcf.analyses.purity.L1PurityAnalysis
...
+    private val evaluationDirCommand = getPlainScallopOption(EvalDirCommand)
+    private val packagesCommand = getPlainScallopOption(PackagesCommand)
+    private val threadsNumCommand = getPlainScallopOption(ThreadsNumCommand)
+    private val analysisNameCommand = getPlainScallopOption(AnalysisNameCommand)
+    private val schedulingStrategyCommand = getPlainScallopOption(SchedulingStrategyCommand)
+
+    verify()
+
+    // Parsed data
+    var classPathFiles = parseCommandWithExternalParser(classPathCommand, ClassPathCommandExternalParser).getOrElse(null)
```
errt: Make all of these `val`s

errt: Don't use `getOrElse`. Probably just leave it as an option and use `get` where the value is actually used?

DEVELOPING_OPAL/tools/src/main/scala/org/opalj/support/info/Purity.scala
```diff
@@ -77,9 +73,139 @@ import org.opalj.tac.fpcf.analyses.purity.L1PurityAnalysis
...
+    var projectDirectory = parseCommandWithInternalParser(projectDirCommand, ProjectDirectoryCommand).getOrElse(null)
+    var libraryDirectory = parseCommandWithInternalParser(libDirCommand, LibraryDirectoryCommand).getOrElse(null)
+    var analysisScheduler = parseCommandWithExternalParser(analysisCommand, AnalysisCommandExternalParser).getOrElse(null)
+    var support: Option[List[FPCFAnalysisScheduler]] = None
+
+    if (fieldAssignabilityCommand.isDefined && escapeCommand.isDefined && eagerCommand.isDefined && analysisScheduler != null) support = Some(parseArgumentsForSupport(
```
errt: Formatting line breaks after if condition and after else

DEVELOPING_OPAL/tools/src/main/scala/org/opalj/support/info/Purity.scala
```diff
@@ -479,235 +577,60 @@ object Purity {
...
-                case "-noJDK"              => withoutJDK = true
-                case "-JDK" =>
-                    cp = JRELibraryFolder; withoutJDK = true
-
-                case unknown =>
-                    Console.println(usage)
-                    throw new IllegalArgumentException(s"unknown parameter: $unknown")
-            }
-            i += 1
+        if(args.contains("--help") || args.contains("-h")) {
```
errt: Can this be made less of a boilerplate? E.g., asking the conf object whether help was asked for?

DEVELOPING_OPAL/tools/src/main/scala/org/opalj/support/info/Purity.scala
```diff
@@ -479,235 +577,60 @@ object Purity {
...
-        if (evaluationDir.isDefined && !evaluationDir.get.exists()) evaluationDir.get.mkdir
-
         val begin = Calendar.getInstance()
         Console.println(begin.getTime)
 
         time {
-            if (multiProjects) {
-                for (subp <- cp.listFiles().filter(_.isDirectory)) {
+            if (purityConf.multiProjects) {
+                for (subp <- purityConf.classPathFiles.head.listFiles().filter(_.isDirectory)) {
```
errt: Using `.head` here (always) seems strange. What would be the semantics if there were multiple files given?

DEVELOPING_OPAL/tools/src/main/scala/org/opalj/support/info/Purity.scala
```diff
@@ -77,9 +73,139 @@ import org.opalj.tac.fpcf.analyses.purity.L1PurityAnalysis
...
+    var jdk: Boolean = parseCommand(jdkCommand).getOrElse(false)
+    var individual: Boolean = parseCommand(individualCommand).getOrElse(false)
+    var closedWorld:Boolean = parseCommand(closedWorldCommand).getOrElse(false)
+    var library: Boolean = parseCommand(libraryCommand).getOrElse(false)
+    var debug: Boolean = parseCommand(debugCommand).getOrElse(false)
+    var multiProjects: Boolean = parseCommand(multiProjectsCommand).getOrElse(false)
+    var evaluationDir = parseCommandWithInternalParser(evaluationDirCommand, EvalDirCommand).getOrElse(null)
+    var packages = parseCommandWithInternalParser(packagesCommand, PackagesCommand).getOrElse(null)
+    var threadsNum: Int = parseCommand(threadsNumCommand).getOrElse(0)
+    var configurationName = parseCommand(analysisNameCommand).getOrElse(null)
```
errt: Using `.getOrElse` here is especially strange since you end up wrapping it into an Option again below.

errt (changes requested):

DEVELOPING_OPAL/tools/src/main/scala/org/opalj/support/info/CallGraph.scala
thanhtung24: This argument configuration class was not done yet and will be deleted

DEVELOPING_OPAL/tools/src/main/scala/org/opalj/support/parser/AnalysisCommandExternalParser.scala
thanhtung24: There is a command called AnalysisCommand which is saved in Common project. This ExternalParser should be used to parse the command while being used in the DEVELOPING_OPAL/Tools project, so I didn't write it as an inner class for purity.

Furthermore, every external parser will be named with the pattern <command name>ExternalParser. That's why this class is called AnalysisCommandExternalParser.

errt: I understand that. But the way it is built now, it does not support anything else but Purity, even looking at the `AnalysisCommand`, which specifies values "L0", "L1", and "L2", which do apply to our purity analyses, but not other analyses in general.

DEVELOPING_OPAL/tools/src/main/scala/org/opalj/support/parser/AnalysisCommandExternalParser.scala
thanhtung24: The genericity of OpalCommandExternalParser helps the compiler to identify the return type of the method OpalConf.parseCommandWithExternalParser.

And yes, the parse method can have an integer or other data types

DEVELOPING_OPAL/tools/src/main/scala/org/opalj/support/parser/AnalysisCommandExternalParser.scala
thanhtung24: This was done for future purposes, if this AnalysisCommand would be used else where in the future, we don't have to move it from Purity, and the AnalysisCommand could just be easily extended. But of course, if it's only used for Purity now, it can be an inner class of purity. Would you prefer the inner class choice?

errt: But how extensible is it? This parser can't be extended like this, because a) it could return purity analyses where other analyses are expected and b) other analyses like FieldAssignability will also use levels like "L2". The `AnalysisCommand` also doesn't seem properly extensible to me, with the predefined levels either not containing levels other analyses would need or containing levels that other analyses wouldn't need.
I think we need to find a better solution here. Another problem that arises is that while many runners have a specific "main" analysis that used to use an argument like "-analysis", many also take arguments for support analyses. And I think there shouldn't be two different commands and parses to support some analysis like FieldAssignability once as "-analysis" and once as, say, "-fieldAssignability"

DEVELOPING_OPAL/tools/src/main/scala/org/opalj/support/parser/AnalysisCommandExternalParser.scala
errt: Then at least move the type parameter to the class, getting rid of the `asInstanceOf`

errt:
Note that there is a compile error in the continuous integration that needs to be fixed

DEVELOPING_OPAL/tools/src/main/scala/org/opalj/support/info/Purity.scala
```diff
@@ -616,16 +639,12 @@ object Purity {
...
-            return
-        }
-
         val begin = Calendar.getInstance()
         Console.println(begin.getTime)
 
         time {
             if (purityConf.multiProjects.get) {
-                for (subp <- purityConf.classPathFiles.get.head.listFiles().filter(_.isDirectory)) {
+                for (subp <- purityConf.classPathFiles.get.flatMap(_.listFiles).filter(_.isDirectory)) {
```
errt: Instead of a flatMap, make this two iterations (see https://docs.scala-lang.org/tour/for-comprehensions.html for an example)

DEVELOPING_OPAL/tools/src/main/scala/org/opalj/support/parser/AnalysisCommandParser.scala
```diff
@@ -0,0 +1,39 @@
...
+ * {{{
+ * val scheduler = AnalysisCommandParser.parse("Purity", "L0")
+ * }}}
+ *
+ * @object AnalysisCommandParser
+ */
+object AnalysisCommandParser {
+    def parse(runnerName: Option[String], analysisLevel: Option[String]): Option[FPCFLazyAnalysisScheduler] =
+        runnerName.get match {
+            case "Purity" =>
```
errt: I don't think this is the right approach. One would have to change this class for every analysis that should be parsed as an analysis command. Instead, we would need a way to defer this parsing to an existing command, e.g. the FieldAssignabilityCommand, EscapeCommand, or, for the case that is now here, a PurityCommand.

DEVELOPING_OPAL/tools/src/main/scala/org/opalj/support/info/Purity.scala
```diff
@@ -78,45 +101,178 @@ import org.opalj.tac.fpcf.analyses.purity.L1PurityAnalysis
...
+
+/**
+ * `PurityConf` is a configuration class for parsing and managing command-line arguments related to purity analysis
+ * in the OPAL framework. It extends `ScallopConf` and mixes in `OpalConf`, providing structured options for various
+ * analysis settings, such as paths, analysis schedulers, and specific analysis configurations.
+ */
+
+class PurityConf(args: Array[String]) extends ScallopConf(args) with OpalConf {
+
+    private object analysis extends Subcommand(AnalysisCommand.name) {
```
errt: This is a lot of boiler plate. Can we get make something general out of this that we then just invoke from here?

errt (changes requested):

OPAL/common/src/main/scala/org/opalj/commandlinebase/OpalConf.scala
```diff
@@ -0,0 +1,62 @@
...
+
+    override def onError(e: Throwable): Unit = e match {
+        case Help("") =>
+            printHelp()
+            sys.exit(0)
+        case Help(_) =>
+            subcommand.foreach(_.printHelp())
+            sys.exit(0)
+        case ScallopException(message) =>
+            println(s"Error: $message")
```
errt: You should still use the OPALLogger facilities to print errors.

OPAL/br/src/main/scala/org/opalj/br/analyses/ProjectsAnalysisApplication.scala
```diff
@@ -0,0 +1,69 @@
...
+    ): Unit = {
+        val (project, report) = analyze(cp, analysisConfig, execution)
+        OPALLogger.log(LogMessage.plainInfo(report.toConsoleString))(project.logContext)
+    }
+
+    protected def analyze(
+        cp:             Iterable[File],
+        analysisConfig: ConfigType,
+        execution:      Int
+    ): (Project[URL], ReportableAnalysisResult) = ???
```
johannesduesing, line 44: Why is this defined as `???` - is this relevant for any scenario?

._command lines
johannesduesing, line 1: What's the purpose of this binary file?

johannesduesing (commented):
I had two small questions when looking through the code. I am also wondering: Where / how are we communicating this new way of invoking OPAL analyses to users? Do we need to change the README / Website?

errt:
Good question. Did we communicate the old way?

._command lines
errt: I don't know how that ended up there, I will remove it.

OPAL/br/src/main/scala/org/opalj/br/analyses/ProjectsAnalysisApplication.scala
errt: Yes. This allows subclasses to not override this method and override the evaluate method instead. If it was abstract (no `???`), that would not be possible. Do you think we should document this?

johannesduesing: Ah okay, i see ... i think we should at least add a code comment or maybe JavaDoc. It would seem counterintuitive to me as a user when i extend this class and call `super.analyze` that i would receive the `an implementation is missing` error. I am also not fully sure why we need both `evaluate` and `analyze`, since the only added functionality (by default) seems to be a `log` statement (and the fact that there is no value returned).

errt: It simplifies implementation for any runners that do not want to produce a `ReportableAnalysisResult`.

johannesduesing: Ah okay, that's fair - then i'm on board with that implementation, especially with the added JavaDoc comment 👍

johannesduesing:
I don't think we did, but maybe because it wasn't that consistent? In my opinion it's a very nice thing to have consistent command line interfaces, we should communicate that at some point 🤓 However i don't think it needs to be part of this PR necessarily.

johannesduesing (dismissed):
LGMT now 👍

johannesduesing (approved):
Latest updates to additional runners also looking good, for me this PR is ready to merge 👍

errt (approved):

Issue #147, opened 2023-10-09
Compiling the java test fixture failed


mohPYdev:
I encountered a compilation error while trying to build the project. Here are the details:

environment: 
- Java 11
- SBT 1.6.2  (increased heap size to 12GB)
- Scala 2.12.15

I followed the instructions mentioned in your README file. the result of " sbt cleanBuild" is the following error:

[log.txt](https://github.com/opalj/opal/files/12842145/log.txt)

Error message:
Compiling the java test fixture failed

errt:
From your log, it is clear that Java 16 is now required to build the test fixtures. We will update the readme accordingly. Thanks for reporting.

Note that we are also using Scala 2.13(.11) now, but that should not be a problem as sbt should download the correct version automatically.

tillbecks:
I encounter the same issue using Java 17, sbt 1.10.(1) and Scala 2.13.15.

errt:
@tillbecks Can you provide your console output?

tillbecks:
Yes of course:
[log.txt](https://github.com/user-attachments/files/17658949/log.txt)

errt:
Your problem seems to be the spaces in your directory path. I can't easily see what encoding the Eclipse BatchCompiler would accept, but you could try replacing the `toString` in line 96 (and probably the implicit one in line 101) of `org.opalj.javacompilation.FixtureCompileSpec` with `toURI`

tillbecks:
Just changing the path did the job for me! Thank you so much

errt:
I thought so, but I would value a pull request to fix the FixtureCompiler not working with paths containing spaces, it is still a bug after all.

PR #232, opened 2024-11-09
Wrap support library paths in quotation marks

errt:
Fixes the second issue recently discovered in #147.

The sourceFolder argument is actually already properly wrapped in quotation marks, but the support libaries were still printed into the command line literally.

Not a huge fan of how the solution works, but it's the best I could find right now.


johannesduesing (approved):
Moved the whole repo to a path with spaces, did a fully `cleanBuild` - everything works as expected 👍

PR #238, opened 2024-11-13
Adapt FPCF IT Tests to new Reachable Methods

johannesduesing:
PRs #237 and #229 lead to callgraphs capturing more methods, which broke the FPCF integration tests. I checked this locally and got a total of 20 failures for the `FPCFAnalysesIntegrationTest`. I went through all of them and confirmed that they are due to new methods being reachable, not old method not being found. I regenerated expected values for these tests and confirmed that the integration tests are working. I'd like to re-run them here on GitHub before merging because i had some resource problems running them locally for two large targets ... 😅


errt:
Something seems to be going on here. You submitted 20 updated files, but the first it:test with the problem had only 14 failing tests. The other it:test runs here then, however, had 22. Either there is something non-deterministic or other factors influenced how many of the tests fail.

errt:
Also, it looks like your /it:test has been skipped, but I don't know why.

johannesduesing:
I think both our improved handling for `Thread.start` and the bugfix for the static initializers lead to new reachable methods - on distinct test cases. That's why it was 14 failing tests after one and then 22 total after merging the other PR.

The 22 vs 20 difference is easily explained - i couldn't run the IT test locally for two JAR files, the scala runtime and OPAl with dependencies due to resource issues 😄 I'm pretty sure they are also failing and need new actual values, i was hoping to confirm that with the runners executing the IT tests. I'll move to our own servers to regenerate them and update the PR!

errt (approved):

PR #239, opened 2024-11-14
Expand check for /it:test workflow to include other author roles

errt:
This should hopefully solve the issue we had recently with the it:test action not running (see #238).


johannesduesing (approved):
LGTM - if i remeber correctly from the last time i worked on this, it needs to be merged to develop before i can test it out, right? I'd use #238 to test this after it's merged 👍

errt:
I think so, too.

johannesduesing:
Works as expected 👍

PR #240, opened 2024-11-15
IDE Basics

Rkoer:
- IDE infrastructure
- Basic IDE solver
- Example implementation of linear constant propagation
- Interaction capabilities
- Example implementation of a linear constant propagation on fields


DEVELOPING_OPAL/validate/src/test/java/org/opalj/fpcf/properties/lcp_on_fields/ArrayValue.java
```diff
@@ -0,0 +1,38 @@
...
+@PropertyValidator(key = LCPOnFieldsProperty.KEY, validator = ArrayValueMatcher.class)
+@Repeatable(ArrayValues.class)
+@Documented
+@Target(ElementType.METHOD)
+@Retention(RetentionPolicy.CLASS)
+public @interface ArrayValue {
+    /**
+     * The name of the variable the array is stored in
+     */
+    String variable();
```
errt: Using (TAC) variable names could be fragile and is hard to understand by reading the source code. Maybe it would be better to identify variables by the line they are declared in.

DEVELOPING_OPAL/validate/src/test/java/org/opalj/fpcf/fixtures/lcp_on_fields/ArrayNativeMethodExample.java
```diff
@@ -0,0 +1,45 @@
+/* BSD 2-Clause License - see OPAL/LICENSE for details. */
+package org.opalj.fpcf.fixtures.lcp_on_fields;
```
errt: Could it be sensible to move these into a subpackage of the fixtures.linear_constant_propagation package?

DEVELOPING_OPAL/validate/src/test/scala/org/opalj/fpcf/ide/IDEPropertiesTest.scala
```diff
@@ -0,0 +1,43 @@
...
+import org.opalj.ide.ConfigKeyDebugLog
+import org.opalj.ide.ConfigKeyTraceLog
+import org.opalj.tac.cg.RTACallGraphKey
+
+/**
+ * Specialized test for IDE analyses preparing the configuration.
+ *
+ * @author Robin Körkemeier
+ */
+class IDEPropertiesTest extends PropertiesTest {
```
errt: This looks like it should be a trait / abstract class

DEVELOPING_OPAL/validate/src/test/scala/org/opalj/fpcf/ide/lcp_on_fields/LCPOnFieldsAnalysisScheduler.scala
```diff
@@ -0,0 +1,17 @@
...
+import org.opalj.tac.fpcf.analyses.ide.instances.lcp_on_fields.LCPOnFieldsAnalysisScheduler
+import org.opalj.tac.fpcf.analyses.ide.integration.JavaIDEAnalysisSchedulerBase
+
+/**
+ * Scheduler for [[org.opalj.tac.fpcf.analyses.ide.instances.lcp_on_fields.problem.LCPOnFieldsProblem]] for Java using
+ * RTA call graph.
+ *
+ * @author Robin Körkemeier
+ */
+object LCPOnFieldsAnalysisScheduler
```
errt: I don't think you need this as a separate object, it should be possible to just put `new LCPOnFieldsAnalysisScheduler() with JavaIDEAnalysisSchedulerBase.RTACallGraph` at its single use. If it is needed, I caution agains naming it the same as its parent class, to avoid confusion.

DEVELOPING_OPAL/validate/src/test/scala/org/opalj/fpcf/properties/AbstractRepeatablePropertyMatcher.scala
```diff
@@ -0,0 +1,70 @@
...
+import org.opalj.br.AnnotationLike
+import org.opalj.br.ObjectType
+import org.opalj.br.analyses.Project
+
+/**
+ * Basic property matcher for repeatable annotations.
+ *
+ * @author Robin Körkemeier
+ */
+abstract class AbstractRepeatablePropertyMatcher extends AbstractPropertyMatcher {
```
errt: This seems like a good idea. Can you apply it to the DirectCallMatcher as well (maybe in a separate PR)?

DEVELOPING_OPAL/validate/src/test/scala/org/opalj/fpcf/properties/lcp_on_fields/LCPOnFieldsMatcher.scala
```diff
@@ -0,0 +1,465 @@
...
+                    val annotation = a.asAnnotationValue.annotation
+                    val expectedFieldName =
+                        getValue(p, constantValueType, annotation.elementValuePairs, "variable").asStringValue.value
+                    val expectedValue =
+                        getValue(p, constantValueType, annotation.elementValuePairs, "value").asIntValue.value
+
+                    (expectedFieldName, expectedValue)
+                }
+
+        val expectedVariableValues =
```
errt: This and `expectedUnknownValues` are practically identical, so extracting a method with the "variableValues"/"unknownValues" as a parameter seems sensible. Could also potentially use this for the `expectedConstantValues`

DEVELOPING_OPAL/validate/src/test/scala/org/opalj/fpcf/properties/lcp_on_fields/LCPOnFieldsMatcher.scala
```diff
@@ -0,0 +1,465 @@
...
+            getValue(p, singleAnnotationType, a.elementValuePairs, "unknownValues").asArrayValue.values
+                .map { a =>
+                    val annotation = a.asAnnotationValue.annotation
+                    val expectedFieldName =
+                        getValue(p, unknownValueType, annotation.elementValuePairs, "variable").asStringValue.value
+
+                    expectedFieldName
+                }
+
+        if (properties.exists {
```
errt: This code of multiple wrapped exists/forall calls with partial functions (i.e., case matches) is hard to understand. Can it be simplified? If not, it should at least be documented what each step does.

DEVELOPING_OPAL/validate/src/test/scala/org/opalj/fpcf/properties/lcp_on_fields/LCPOnFieldsMatcher.scala
errt: Note that the `ArrayValueMatcher` and `StaticValuesMatcher` classes also use duplicated code to be extracted to a common base trait/class. Maybe an LCPOnFieldsMatcher trait to follow the file name.

OPAL/tac/src/main/scala/org/opalj/tac/fpcf/analyses/ide/solver/JavaForwardICFG.scala
```diff
@@ -0,0 +1,55 @@
...
+import scala.collection
+import scala.collection.immutable
+import scala.collection.mutable
+
+import org.opalj.br.Method
+import org.opalj.br.analyses.SomeProject
+
+/**
+ * Interprocedural control flow graph for Java programs in forward direction. This implementation is based on the
+ * [[org.opalj.tac.fpcf.analyses.ifds.JavaForwardICFG]] from IFDS.
```
errt, line 17: Would it be possible to merge these implementations? It seems like a bad idea to maintain two very similar implementations when both IFDS and IDE could probably use the same

OPAL/ide/src/main/scala/org/opalj/ide/solver/ICFG.scala
```diff
@@ -0,0 +1,58 @@
...
+     */
+    def getCallable(stmt: Statement): Callable
+
+    /**
+     * Build a string representation of a statement. Is only used for debugging purposes.
+     *
+     * @param indent to use on newlines (e.g. indentation for prettier logs)
+     * @param short whether to build a long or a more compact string
+     */
+    def stringifyStatement(stmt: Statement, indent: String = "", short: Boolean = false): String
```
errt: It seems weird to have this on the ICFG. Should it be on the statement (if needed at all)?

OPAL/tac/src/main/scala/org/opalj/tac/fpcf/analyses/ide/solver/JavaBaseICFG.scala
```diff
@@ -0,0 +1,103 @@
...
+                    }
+                    .toSet
+            case _ =>
+                throw new IllegalStateException("Call graph must be computed before the analysis starts!")
+        }
+    }
+
+    override def getCallable(javaStmt: JavaStatement): Method = javaStmt.method
+
+    override def stringifyStatement(javaStmt: JavaStatement, indent: String = "", short: Boolean = false): String = {
```
errt: Is this even used with `short=false` or any indent anywhere? It seems that this method could be removed completely or moved to the `FlowRecordingIDEProblem`

OPAL/tac/src/main/scala/org/opalj/tac/fpcf/analyses/ide/solver/JavaBaseICFG.scala
```diff
@@ -0,0 +1,103 @@
...
+import org.opalj.br.fpcf.PropertyStoreKey
+import org.opalj.br.fpcf.analyses.ContextProvider
+import org.opalj.br.fpcf.properties.cg.Callees
+import org.opalj.fpcf.FinalP
+import org.opalj.fpcf.PropertyStore
+import org.opalj.value.ValueInformation
+
+/**
+ * Base interprocedural control flow graph for Java programs. This implementation is based on the
+ * [[org.opalj.tac.fpcf.analyses.ifds.JavaICFG]] from IFDS.
```
errt, line 26: See below, can this be merged with the IFDS variant to avoid duplication?

OPAL/tac/src/main/scala/org/opalj/tac/fpcf/analyses/ide/integration/JavaIDEAnalysisSchedulerBase.scala
```diff
@@ -0,0 +1,80 @@
...
+
+    override def uses: Set[PropertyBounds] =
+        super.uses.union(immutable.Set(
+            PropertyBounds.finalP(TACAI),
+            PropertyBounds.finalP(Callers)
+        ))
+}
+
+object JavaIDEAnalysisSchedulerBase {
+    /**
```
errt, line 49: I don't understand the point of these traits

OPAL/tac/src/main/scala/org/opalj/tac/fpcf/analyses/ide/integration/JavaIDEAnalysisSchedulerBase.scala
```diff
@@ -0,0 +1,80 @@
...
+     * Trait to drop-in [[org.opalj.tac.cg.RTACallGraphKey]] as [[callGraphKey]]
+     */
+    trait RTACallGraph {
```
errt: This should not be tied to RTA in particular, but allow users to pick a call graph algorithm of their choice.

OPAL/tac/src/main/scala/org/opalj/tac/fpcf/analyses/ide/integration/JavaIDEAnalysisSchedulerBase.scala
```diff
@@ -0,0 +1,80 @@
...
+
+    override def requiredProjectInformation: ProjectInformationKeys =
+        super.requiredProjectInformation ++ Seq(
+            DeclaredMethodsKey,
+            TypeIteratorKey,
+            callGraphKey
+        )
+
+    override def uses: Set[PropertyBounds] =
+        super.uses.union(immutable.Set(
```
errt: Use `++` or similar instead of `union`

OPAL/tac/src/main/scala/org/opalj/tac/fpcf/analyses/ide/integration/JavaIDEAnalysisSchedulerBase.scala
```diff
@@ -0,0 +1,80 @@
...
+ * A base IDE analysis scheduler for Java programs.
+ *
+ * @author Robin Körkemeier
+ */
+abstract class JavaIDEAnalysisSchedulerBase[Fact <: IDEFact, Value <: IDEValue]
+    extends IDEAnalysisScheduler[Fact, Value, JavaStatement, Method, JavaICFG] {
+    /**
+     * Key indicating which call graph should be used
+     */
+    val callGraphKey: CallGraphKey
```
errt: Is this really needed?

OPAL/tac/src/main/scala/org/opalj/tac/fpcf/analyses/ide/integration/JavaIDEAnalysisSchedulerBase.scala
```diff
@@ -0,0 +1,80 @@
...
+
+    override def requiredProjectInformation: ProjectInformationKeys =
+        super.requiredProjectInformation ++ Seq(
+            DeclaredMethodsKey,
+            TypeIteratorKey,
+            callGraphKey
```
errt: Also, is this really needed? CallGraphKeys are meant to make it easier to compute a call graph, but their use should not be mandated by analyses. Instead, having `Callers` (and/or `Callees`, depending on what the analysis actually uses) should suffice.

OPAL/tac/src/main/scala/org/opalj/tac/fpcf/analyses/ide/instances/linear_constant_propagation/problem/LinearConstantPropagationLattice.scala
```diff
@@ -0,0 +1,34 @@
...
+object LinearConstantPropagationLattice extends MeetLattice[LinearConstantPropagationValue] {
+    override def top: LinearConstantPropagationValue = UnknownValue
+
+    override def bottom: LinearConstantPropagationValue = VariableValue
+
+    override def meet(
+        x: LinearConstantPropagationValue,
+        y: LinearConstantPropagationValue
+    ): LinearConstantPropagationValue = (x, y) match {
+        case (UnknownValue, y)                                  => y
```
errt: This shadows the `y` parameter with a new `y`. Either use a new name or just use `_` here, since the `y` parameter is still available.
Also, use more descriptive names than x and y.

OPAL/ide/src/main/scala/org/opalj/ide/problem/MeetLattice.scala
```diff
@@ -0,0 +1,26 @@
...
+package org.opalj
+package ide
+package problem
+
+/**
+ * Interface representing the lattice that orders the IDE values.
+ *
+ * @author Robin Körkemeier
+ */
+trait MeetLattice[Value <: IDEValue] {
```
errt, line 11: Why is this a separate trait? Usually, we'd include things like the meet operator within the property class itself.

OPAL/tac/src/main/scala/org/opalj/tac/fpcf/analyses/ide/instances/lcp_on_fields/problem/LCPOnFieldsValue.scala
```diff
@@ -0,0 +1,69 @@
...
+
+/**
+ * Value representing the value of a static field.
+ *
+ * @author Robin Körkemeier
+ */
+case class StaticFieldValue(
+    value: LinearConstantPropagationValue
+) extends LCPOnFieldsValue {
+    override def toString: String = s"StaticFieldValue($value)"
```
errt: Is this necessary? Cases classes have a default toString

OPAL/tac/src/main/scala/org/opalj/tac/fpcf/analyses/ide/instances/lcp_on_fields/LinearConstantPropagationAnalysisSchedulerExtended.scala
```diff
@@ -0,0 +1,48 @@
...
+import org.opalj.tac.fpcf.analyses.ide.instances.linear_constant_propagation.problem.LinearConstantPropagationFact
+import org.opalj.tac.fpcf.analyses.ide.instances.linear_constant_propagation.problem.LinearConstantPropagationValue
+import org.opalj.tac.fpcf.analyses.ide.integration.JavaIDEAnalysisScheduler
+import org.opalj.tac.fpcf.analyses.ide.integration.JavaIDEAnalysisSchedulerBase
+import org.opalj.tac.fpcf.analyses.ide.integration.JavaIDEPropertyMetaInformation
+import org.opalj.tac.fpcf.analyses.ide.problem.JavaIDEProblem
+import org.opalj.tac.fpcf.analyses.ide.solver.JavaICFG
+
+/**
+ * Extended linear constant propagation as IDE analysis.
```
errt: What does `extended` mean here? Use a more descriptive name and improve the documentation.

OPAL/tac/src/main/scala/org/opalj/tac/fpcf/analyses/ide/instances/lcp_on_fields/LCPOnFieldsAnalysisScheduler.scala
```diff
@@ -0,0 +1,55 @@
...
+        new LCPOnFieldsProblem(project, icfg)
+    }
+
+    override def requiredProjectInformation: ProjectInformationKeys =
+        super.requiredProjectInformation ++ Seq(
+            DeclaredFieldsKey
+        )
+
+    override def uses: Set[PropertyBounds] =
+        super.uses.union(immutable.Set(
```
errt: Use `++` or similar instead of `union`

OPAL/tac/src/main/scala/org/opalj/tac/fpcf/analyses/ide/instances/lcp_on_fields/LCPOnFieldsAnalysisScheduler.scala
```diff
@@ -0,0 +1,55 @@
...
+
+    override def createProblem(project: SomeProject, icfg: JavaICFG): JavaIDEProblem[
+        LCPOnFieldsFact,
+        LCPOnFieldsValue
+    ] = {
+        new LCPOnFieldsProblem(project, icfg)
+    }
+
+    override def requiredProjectInformation: ProjectInformationKeys =
+        super.requiredProjectInformation ++ Seq(
```
errt: If you only add one, avoid creating an additional Seq and use `:+` instead.

OPAL/tac/src/main/scala/org/opalj/tac/fpcf/analyses/ide/instances/lcp_on_fields/LinearConstantPropagationAnalysisSchedulerExtended.scala
```diff
@@ -0,0 +1,48 @@
...
+
+    override def createProblem(project: SomeProject, icfg: JavaICFG): JavaIDEProblem[
+        LinearConstantPropagationFact,
+        LinearConstantPropagationValue
+    ] = {
+        new LinearConstantPropagationProblemExtended
+    }
+
+    override def uses: Set[PropertyBounds] =
+        super.uses.union(immutable.Set(
```
errt: Use `:+` or similar instead of `union`

OPAL/ide/src/main/scala/org/opalj/ide/util/Logging.scala
```diff
@@ -0,0 +1,81 @@
...
+package ide
+package util
+
+import org.opalj.br.analyses.SomeProject
+import org.opalj.log.GlobalLogContext
+import org.opalj.log.LogContext
+import org.opalj.log.OPALLogger
+
+/**
+ * Logging extension for IDE.
```
errt: Is this really needed? And is there anything IDE specific to it?

OPAL/ide/src/main/scala/org/opalj/ide/solver/IDEAnalysisProxy.scala
```diff
@@ -0,0 +1,139 @@
...
+import org.opalj.ide.integration.IDEPropertyMetaInformation
+import org.opalj.ide.integration.IDETargetCallablesProperty
+import org.opalj.ide.problem.IDEFact
+import org.opalj.ide.problem.IDEValue
+import org.opalj.ide.util.Logging
+
+/**
+ * A proxy for IDE analyses that accepts analysis requests for callables as well as statement-callable combinations.
+ * The [[IDEAnalysis]] solver runs on callables only and additionally produces results for each statement of that
+ * callable. This proxy analysis reduces all analysis requests to the callable and then forward it to the actual IDE
```
errt: `forwards them`?

OPAL/ide/src/main/scala/org/opalj/ide/solver/IDEAnalysisProxy.scala
```diff
@@ -0,0 +1,139 @@
...
+class IDEAnalysisProxy[Fact <: IDEFact, Value <: IDEValue, Statement, Callable <: Entity](
+    val project:                 SomeProject,
+    val propertyMetaInformation: IDEPropertyMetaInformation[Fact, Value, Statement, Callable]
+) extends FPCFAnalysis with Logging.ByProjectConfig {
+    /**
+     * @param entity either only a callable or a pair of callable and statement that should be analyzed (if no statement
+     *               is given, the result for all exit statements is calculated)
+     */
+    def proxyAnalysis(entity: Entity): ProperPropertyComputationResult = {
+        logInfo(s"proxying request to ${PropertyKey.name(propertyMetaInformation.key)} for $entity")
```
errt: Is this necessary?

OPAL/ide/src/main/scala/org/opalj/ide/problem/FlowRecordingIDEProblem.scala
```diff
@@ -0,0 +1,381 @@
...
+    /**
+     * A node in the graph is the combination of a statement and a fact.
+     */
+    val NODE_AS_STMT_AND_FACT: FlowRecorderModes.Value = Value
+}
+
+/**
+ * Wrapper class for a normal IDE problem for debugging purposes. Records the flow paths the IDE solver takes for a
+ * given base problem as graph and writes it to a file in DOT format.
+ * DOT files can either be viewed with a suitable local program or online e.g. at
```
errt: I don't think an explanation on how to view DOT files is necessary here

OPAL/ide/src/main/scala/org/opalj/ide/problem/EdgeFunction.scala
```diff
@@ -0,0 +1,103 @@
...
+    override def meetWith(otherEdgeFunction: EdgeFunction[Value]): EdgeFunction[Value] = {
+        if (otherEdgeFunction.equalTo(this)) {
+            this
+        } else {
+            otherEdgeFunction.meetWith(this)
+        }
+    }
+
+    override def equalTo(otherEdgeFunction: EdgeFunction[Value]): Boolean =
+        (otherEdgeFunction eq this) || otherEdgeFunction.isInstanceOf[IdentityEdgeFunction[Value]]
```
errt: Don't do `isInstanceOf` (and `asInstanceOf`, below), but use pattern matching if possible.

OPAL/ide/src/main/scala/org/opalj/ide/problem/EdgeFunction.scala
```diff
@@ -0,0 +1,103 @@
...
+
+    /**
+     * Combine two edge functions via meet semantics
+     */
+    def meetWith(otherEdgeFunction: EdgeFunction[Value]): EdgeFunction[Value]
+
+    /**
+     * Check whether two edge functions are equal (s.t. they produce the same result for same source values)
+     */
+    def equalTo(otherEdgeFunction: EdgeFunction[Value]): Boolean
```
errt: Why isn't this just `equals`?

OPAL/ide/src/main/scala/org/opalj/ide/problem/EdgeFunction.scala
```diff
@@ -0,0 +1,103 @@
...
+     * @param secondEdgeFunction the edge function that is applied after this one
+     * @return an edge function computing the same values as first applying this edge function and then applying the
+     *         result to the second edge function
+     */
+    def composeWith(secondEdgeFunction: EdgeFunction[Value]): EdgeFunction[Value]
+
+    /**
+     * Combine two edge functions via meet semantics
+     */
+    def meetWith(otherEdgeFunction: EdgeFunction[Value]): EdgeFunction[Value]
```
errt: Why isn't this just `meet`?

OPAL/ide/src/main/scala/org/opalj/ide/integration/IDERawProperty.scala
```diff
@@ -0,0 +1,53 @@
...
+
+import org.opalj.fpcf.Property
+import org.opalj.fpcf.PropertyKey
+import org.opalj.ide.problem.IDEFact
+import org.opalj.ide.problem.IDEValue
+
+/**
+ * Class representing a property that is directly created by an IDE analysis.
+ *
+ * @param key the property key (very likely taken from an [[IDERawPropertyMetaInformation]] instance)
```
errt: `very likely` doesn't sound very reassuring.

OPAL/ide/src/main/scala/org/opalj/ide/integration/FlowRecordingAnalysisScheduler.scala
```diff
@@ -0,0 +1,186 @@
...
+import org.opalj.ide.problem.IDEProblem
+import org.opalj.ide.problem.IDEValue
+import org.opalj.ide.solver.ICFG
+import org.opalj.ide.solver.IDEAnalysis
+import org.opalj.ide.util.Logging
+
+/**
+ * Wrapper class for a normal IDE analysis scheduler for debugging purposes. Records the flow paths the IDE solver takes
+ * for a given base problem as graph and writes it to a file in DOT format.
+ * DOT files can either be viewed with a suitable local program or online e.g. at
```
errt: See below, explanation on DOT files probably not necessary

DEVELOPING_OPAL/validate/src/test/scala/org/opalj/fpcf/ide/lcp_on_fields/LCPOnFieldsTests.scala
```diff
@@ -0,0 +1,53 @@
...
+ * Test runner for linear constant propagation on fields.
+ *
+ * @author Robin Körkemeier
+ */
+class LCPOnFieldsTests extends IDEPropertiesTest {
+    override def fixtureProjectPackage: List[String] = {
+        List("org/opalj/fpcf/fixtures/lcp_on_fields")
+    }
+
+    describe("Execute the o.o.t.f.a.i.i.l.LCPOnFieldsAnalysis") {
```
errt: Can probably live without the abbreviated packages

OPAL/tac/src/main/scala/org/opalj/tac/fpcf/analyses/ide/instances/lcp_on_fields/problem/LinearConstantPropagationProblemExtended.scala
```diff
@@ -0,0 +1,280 @@
...
+        }
+    }
+
+    private def getObjectFieldFromProperty(
+        objectVar: JavaStatement.V,
+        fieldName: String
+    )(property: LCPOnFieldsPropertyMetaInformation.Self): LinearConstantPropagationValue = {
+        property
+            .results
+            .filter {
```
errt: Consider using `.collect` instead of `.filter{ ... }.map{ ... }` as it is more efficient. Alternatively, it might be possible to integrate the `map` (or even the `filter`) into the `foldLeft`. Also consider using `fold` instead of `foldLeft` if possible.

OPAL/tac/src/main/scala/org/opalj/tac/fpcf/analyses/ide/instances/lcp_on_fields/problem/LinearConstantPropagationProblemExtended.scala
```diff
@@ -0,0 +1,280 @@
...
+        }
+    }
+
+    private def getStaticFieldFromProperty(
+        objectType: ObjectType,
+        fieldName:  String
+    )(property: LCPOnFieldsPropertyMetaInformation.Self): LinearConstantPropagationValue = {
+        property
+            .results
+            .filter {
```
errt: See above on replacing filter+map

OPAL/tac/src/main/scala/org/opalj/tac/fpcf/analyses/ide/instances/lcp_on_fields/problem/LinearConstantPropagationProblemExtended.scala
```diff
@@ -0,0 +1,280 @@
...
+        }
+    }
+
+    private def getArrayElementFromProperty(
+        arrayVar: JavaStatement.V,
+        index:    Option[Int]
+    )(property: LCPOnFieldsPropertyMetaInformation.Self): LinearConstantPropagationValue = {
+        property
+            .results
+            .filter {
```
errt: See below on replacing filter+map

OPAL/tac/src/main/scala/org/opalj/tac/fpcf/analyses/ide/instances/lcp_on_fields/problem/LinearConstantPropagationProblemExtended.scala
```diff
@@ -0,0 +1,280 @@
...
+        targetFact: LinearConstantPropagationFact
+    )(implicit propertyStore: PropertyStore): EdgeFunctionResult[LinearConstantPropagationValue] = {
+        if (sourceFact == nullFact) {
+            return UnknownValueEdgeFunction
+        }
+
+        val arrayVar = arrayLoadExpr.arrayRef.asVar
+
+        val index = arrayLoadExpr.index.asVar.value match {
+            case intRange: DefaultIntegerRangeValues#IntegerRange =>
```
errt: Check if the type projection is necessary. OPAL is about to switch to Scala 3 which has dropped general type projection.

OPAL/tac/src/main/scala/org/opalj/tac/fpcf/analyses/ide/instances/linear_constant_propagation/problem/LinearConstantPropagationProblem.scala
```diff
@@ -0,0 +1,641 @@
...
+
+    override val lattice: MeetLattice[LinearConstantPropagationValue] =
+        LinearConstantPropagationLattice
+
+    override def getAdditionalSeeds(stmt: JavaStatement, callee: Method)(
+        implicit propertyStore: PropertyStore
+    ): collection.Set[LinearConstantPropagationFact] = {
+        callee.parameterTypes
+            .zipWithIndex
+            .filter { case (paramType, _) => paramType.isIntegerType }
```
errt: Consider replacing filter+map with collect, which is more efficient.

OPAL/tac/src/main/scala/org/opalj/tac/fpcf/analyses/ide/instances/linear_constant_propagation/problem/LinearConstantPropagationProblem.scala
```diff
@@ -0,0 +1,641 @@
...
+    override val nullFact: LinearConstantPropagationFact =
+        NullFact
+
+    override val lattice: MeetLattice[LinearConstantPropagationValue] =
+        LinearConstantPropagationLattice
+
+    override def getAdditionalSeeds(stmt: JavaStatement, callee: Method)(
+        implicit propertyStore: PropertyStore
+    ): collection.Set[LinearConstantPropagationFact] = {
+        callee.parameterTypes
```
errt, line 44: Consider placing a `.iterator` here to avoid repeatedly creating temporary collections

OPAL/ide/src/main/scala/org/opalj/ide/problem/EdgeFunction.scala
```diff
@@ -0,0 +1,103 @@
...
+     */
+    def equalTo(otherEdgeFunction: EdgeFunction[Value]): Boolean
+}
+
+/**
+ * Special edge function representing an identity edge function.
+ *
+ * @author Robin Körkemeier
+ */
+case class IdentityEdgeFunction[Value <: IDEValue]() extends EdgeFunction[Value] {
```
errt: This should probably be a (case) object

OPAL/tac/src/main/scala/org/opalj/tac/fpcf/analyses/ide/instances/lcp_on_fields/problem/LCPOnFieldsEdgeFunctions.scala
```diff
@@ -0,0 +1,510 @@
...
+                        fieldName2 -> linear_constant_propagation.problem.VariableValue
+                    } +
+                        (fieldName -> LinearConstantPropagationLattice.meet(
+                            value,
+                            values.getOrElse(fieldName, linear_constant_propagation.problem.VariableValue)
+                        ))
+                )
+
+            case IdentityEdgeFunction()   => this
+            case AllTopEdgeFunction(_)    => this
```
errt: If you don't need to match the parameters, you can replace cases like this with, e.g., `case _: AllTopEdgeFunction`

OPAL/tac/src/main/scala/org/opalj/tac/fpcf/analyses/ide/instances/lcp_on_fields/problem/LCPOnFieldsEdgeFunctions.scala
```diff
@@ -0,0 +1,510 @@
...
+            })
+}
+
+/**
+ * Edge function for initializing an object.
+ *
+ * @author Robin Körkemeier
+ */
+object NewObjectEdgeFunction extends ObjectEdgeFunction(immutable.Map.empty) {
+    override def toString: String = "NewObjectEdgeFunction()"
```
errt, line 123: You can probably get rid of toString methods like this if you just make these objects `case object`s instead, which provide a default toString method

OPAL/tac/src/main/scala/org/opalj/tac/fpcf/analyses/ide/instances/lcp_on_fields/problem/LCPOnFieldsProblem.scala
```diff
@@ -0,0 +1,831 @@
...
+        (if (callee.isStatic) {
+             immutable.Set.empty
+         } else {
+             /* Add fact for `this` */
+             immutable.Set(ObjectFact("param0", -1))
+         }) ++
+            /* Add facts for other parameters */
+            callee.parameterTypes
+                .zipWithIndex
+                .filter { case (paramType, _) => paramType.isObjectType || paramType.isArrayType }
```
errt: Consider replacing filter+map with collect, plus performing the whole operation on an iterator

OPAL/tac/src/main/scala/org/opalj/tac/fpcf/analyses/ide/instances/lcp_on_fields/problem/LCPOnFieldsProblem.scala
```diff
@@ -0,0 +1,831 @@
...
+                        ObjectFact(s"param${index + 1}", -(index + 2))
+                    } else {
+                        ArrayFact(s"param${index + 1}", -(index + 2))
+                    }
+                }
+                .toSet ++
+            /* Add facts for static fields of class */
+            callee.classFile
+                .fields
+                .filter(_.isStatic)
```
errt: Again, consider collect + iterator

OPAL/tac/src/main/scala/org/opalj/tac/fpcf/analyses/ide/instances/lcp_on_fields/problem/LCPOnFieldsProblem.scala
```diff
@@ -0,0 +1,831 @@
...
+                                            return linear_constant_propagation.problem.VariableValue
+                                    }
+                                }
+
+                            case _ =>
+                        }
+                    }
+
+                    seenStmts.addAll(workingStmts)
+                    workingStmts = workingStmts.foldLeft(mutable.Set.empty[JavaStatement]) { (nextStmts, stmt) =>
```
errt: Consider using fold over foldLeft if possible

OPAL/tac/src/main/scala/org/opalj/tac/fpcf/analyses/ide/instances/lcp_on_fields/problem/LCPOnFieldsProblem.scala
```diff
@@ -0,0 +1,831 @@
...
+        NullFact
+
+    override val lattice: MeetLattice[LCPOnFieldsValue] =
+        LCPOnFieldsLattice
+
+    override def getAdditionalSeeds(stmt: JavaStatement, callee: Method)(
+        implicit propertyStore: PropertyStore
+    ): collection.Set[LCPOnFieldsFact] = {
+        (if (callee.isStatic) {
+             immutable.Set.empty
```
errt: It should be possible to get around all these `immutable.Set` by importing immutable.Set directly and importing mutable.Set with renaming, i.e., `import scala.collections.mutable.{Set => MutableSet}`

OPAL/ide/src/main/scala/org/opalj/ide/solver/IDEAnalysis.scala
```diff
@@ -0,0 +1,1115 @@
...
+                        (
+                            n,
+                            dValuePairs.groupMapReduce(_._1)(_._2) { (value1, value2) =>
+                                problem.lattice.meet(value1, value2)
+                            }.toSet
+                        )
+                    }
+
+                val resultsForExit = resultsByStatement
+                    .filter { case (n, _) => icfg.isNormalExitStatement(n) }
```
errt: Consider replacing filter+map with collect

OPAL/ide/src/main/scala/org/opalj/ide/solver/IDEAnalysis.scala
```diff
@@ -0,0 +1,1115 @@
...
+        }
+
+        def addDependee(eOptionP: SomeEOptionP, c: () => Unit): Unit = {
+            // The eOptionP is only inserted the first time the corresponding EPK occurs. Consequently, it is the most
+            // precise property result that is seen by all dependents.
+            val (_, set) = dependees.getOrElseUpdate(eOptionP.toEPK, (eOptionP, mutable.Set.empty))
+            set.add(c)
+        }
+
+        def areDependeesEmpty: Boolean = {
```
errt: It would be more conventional to have a `hasDependees` method with inverted semantics

OPAL/ide/src/main/scala/org/opalj/ide/solver/IDEAnalysis.scala
```diff
@@ -0,0 +1,1115 @@
...
+        performPhase1()
+        performPhase2()
+
+        createResult()
+    }
+
+    /**
+     * @return whether the phase is finished or has to be continued once the dependees are resolved
+     */
+    private def performPhase1()(implicit s: State): Boolean = {
```
errt, line 523: Phase1/Phase2 are very non-descriptive names. Try to come up with better ones and at the very least, document these methods properly to make the code understandable.

OPAL/ide/src/main/scala/org/opalj/ide/solver/IDEAnalysis.scala
```diff
@@ -0,0 +1,1115 @@
...
+                    if (eOptionP.hasUBP && eOptionP.ub == ideRawProperty) {
+                        None
+                    } else {
+                        Some(InterimEUBP(callable, ideRawProperty))
+                    }
+                }
+            )
+        }
+
+        logDebug("finished creation of properties and results")
```
errt: Consider whether these log messages are really necessary

OPAL/ide/src/main/scala/org/opalj/ide/solver/IDEAnalysis.scala
```diff
@@ -0,0 +1,1115 @@
...
+                processExitFlow(path, f)
+            } else { // IDE P1 line 30
+                processNormalFlow(path, f)
+            }
+
+            logDebug(s"${s.getPathWorkListSize} path(s) remaining after processing last path")
+        }
+    }
+
+    private def processCallFlow(path: Path, f: JumpFunction, qs: collection.Set[Callable])(
```
errt: This code is poorly documented and uses many nondescriptives names. What are things like `sp`, `n`, `rs`, `qs`, `d5`, `d5s`, `d3`, `sqs`, etc.

OPAL/ide/src/main/scala/org/opalj/ide/solver/IDEAnalysis.scala
```diff
@@ -0,0 +1,1115 @@
...
+import org.opalj.ide.util.Logging
+
+/**
+ * Basic solver for IDE problems. Uses the exhaustive algorithm that was presented in the original IDE paper from 1996
+ * as base. For an example problem have a look at `LinearConstantPropagationProblem` in the TAC module. For an example
+ * of interacting IDE problems have a look at `LCPOnFieldsProblem` and `LinearConstantPropagationProblemExtended`.
+ *
+ * @author Robin Körkemeier
+ */
+class IDEAnalysis[Fact <: IDEFact, Value <: IDEValue, Statement, Callable <: Entity](
```
errt, line 59: There are many complex methods in this class that should be properly documented, at least with a ScalaDoc on what they do.

errt (changes requested):

DEVELOPING_OPAL/validate/src/test/java/org/opalj/fpcf/properties/lcp_on_fields/ArrayValue.java
Rkoer: I guess we could add this, but there are tests (not the ones for LCPoF but esp. the ones for LCP) are more detailed and also check values that are not explicitly declared in the code.

DEVELOPING_OPAL/validate/src/test/java/org/opalj/fpcf/fixtures/lcp_on_fields/ArrayNativeMethodExample.java
Rkoer: That sounds logical. However, at the moment there are two separate test classes, one for LCP and one for LCPoF. When moving the fixtures inside the linear_constant_propagation package then the tests for LCP do pick up the fixtures for LCPoF too. I guess we would need to specify each test fixture for LCP manually instead of providing the package. Is there an alternative solution to this?

DEVELOPING_OPAL/validate/src/test/scala/org/opalj/fpcf/properties/AbstractRepeatablePropertyMatcher.scala
Rkoer: Sure, see PR#274

DEVELOPING_OPAL/validate/src/test/java/org/opalj/fpcf/properties/lcp_on_fields/ArrayValue.java
errt: Can we identify them by program counter or line number then? Program counters are at least guaranteed to be stable since we use a fixed compiler. Line numbers are a little more fragile as one might make non-semantic changes to the source files, but they are easier to understand from the source code.

DEVELOPING_OPAL/validate/src/test/java/org/opalj/fpcf/fixtures/lcp_on_fields/ArrayNativeMethodExample.java
errt: How about one fixture directory with two subdirectories, for LCP and LCPoF?

OPAL/tac/src/main/scala/org/opalj/tac/fpcf/analyses/ide/solver/JavaForwardICFG.scala
Rkoer: The IDE variant relies on separate call-site and return-site nodes which do not exist in the IFDS variant. It also deals with call statements differently (e.g. the IDE solver does not allow for call statements without call targets while the IFDS solver treats such call statments as normal statements). Merging the implementations seems possible to me but requires changes to the IFDS solver and potentially breaks existing IFDS problem specifications.

OPAL/tac/src/main/scala/org/opalj/tac/fpcf/analyses/ide/solver/JavaBaseICFG.scala
Rkoer: I used the stringification of statements in the solver for debugging (in general there are many `logDebug` and `logTrace` statments i used to understand and debug the solver). I guess in one of the following comments you mentioned to remove some of the log statements. Do you want me to remove the logging (or maybe just the debug/trace messages) from the solver entirely? In this case i would remove the `stringifyStatement` also instead of refactoring it.

OPAL/ide/src/main/scala/org/opalj/ide/util/Logging.scala
Rkoer: It is not in particular IDE specific. I used debug and trace messages to debug the solver. OPAL did only offer warn, info and debug messages. Differentiating between debug and trace allowed me to log more information only if debug information did not suffice to find bugs.

OPAL/tac/src/main/scala/org/opalj/tac/fpcf/analyses/ide/integration/JavaIDEAnalysisSchedulerBase.scala
Rkoer: It allows to build analysis schedulers by using mixins (e.g. `with JavaIDEAnalysisSchedulerBase.ForwardICFG`and `with JavaIDEAnalysisSchedulerBase.RTACallGraph`) instead of implementing the corresponding methods fully by hand in the analysis scheduler.

OPAL/tac/src/main/scala/org/opalj/tac/fpcf/analyses/ide/integration/JavaIDEAnalysisSchedulerBase.scala
Rkoer: One has to write a separate trait per call graph algorithm to use this scheme of configuring the analysis scheduler. It is not required to use JavaIDEAnalysisSchedulerBase. One can as well implement the methods that these traits provide by hand in the analysis scheduler. I just considered it a convenient way to build IDE analysis schedulers and abstracting over the detailed implementation of how to configure the call graph key (and the ICFG).

OPAL/tac/src/main/scala/org/opalj/tac/fpcf/analyses/ide/integration/JavaIDEAnalysisSchedulerBase.scala
Rkoer: Ah, now i get what you mean. I guess dependencies to Callers is already there. I removed the `callGraphKey` property as well as the `RTACallGraphKey` trait.

OPAL/ide/src/main/scala/org/opalj/ide/problem/MeetLattice.scala
Rkoer: I decided to model this separately and closer to the IDE problem definition because it is the lattice for IDE values. The meet-implementations i have seen in the code of other OPAL analyses are rather about meeting the resulting analysis properties. For IDE this would include the IDE fact in addition to the IDE value. But i think i could move it into the property definition if you want me to.

OPAL/ide/src/main/scala/org/opalj/ide/util/Logging.scala
errt: In this case, maybe expand the general logging facilities with trace messages instead of adding this only in the IDE subproject?

OPAL/tac/src/main/scala/org/opalj/tac/fpcf/analyses/ide/solver/JavaBaseICFG.scala
errt: In general, yes, I'd suggest to remove most logging. Messages (of the appropriate logging levels) should primarily be used to notify users of things that might affect their analysis, not to provide information into the analysis process itself.

OPAL/tac/src/main/scala/org/opalj/tac/fpcf/analyses/ide/solver/JavaForwardICFG.scala
errt: There are not many existing IFDS problems, so merging now would probably be a good idea before more client analyses exist and implementations diverge even more.

DEVELOPING_OPAL/validate/src/test/scala/org/opalj/fpcf/properties/ide/IDEPropertyMatcherMixin.scala
```diff
@@ -0,0 +1,162 @@
...
+import org.opalj.fpcf.Property
+import org.opalj.fpcf.properties.AbstractPropertyMatcher
+import org.opalj.ide.integration.BasicIDEProperty
+
+/**
+ * An [[AbstractPropertyMatcher]] with additional functions useful to write property matchers for IDE analyses.
+ *
+ * @author Robin Körkemeier
+ */
+trait IDEPropertyMatcherMixin extends AbstractPropertyMatcher {
```
errt: Remove `Mixin` from the name

DEVELOPING_OPAL/validate/src/test/java/org/opalj/fpcf/properties/linear_constant_propagation/lcp/ConstantValue.java
```diff
@@ -18,9 +18,9 @@
 @Retention(RetentionPolicy.CLASS)
 public @interface ConstantValue {
     /**
-     * The name of the variable
+     * The PC of this variable in the TAC
```
errt: PCs are not a property of the TAC, but of the bytecode. Do not confuse them with TAC indices.

OPAL/ide/src/main/scala/org/opalj/ide/problem/EdgeFunction.scala
```diff
@@ -57,7 +57,11 @@ case class IdentityEdgeFunction[Value <: IDEValue]() extends EdgeFunction[Value]
     }
 
     override def equalTo(otherEdgeFunction: EdgeFunction[Value]): Boolean =
-        (otherEdgeFunction eq this) || otherEdgeFunction.isInstanceOf[IdentityEdgeFunction[Value]]
+        (otherEdgeFunction eq this) ||
+            (otherEdgeFunction match {
+                case IdentityEdgeFunction() => true
```
errt: Better: `case _: IdentityEdgeFuction`

But given this case class has no fields, shouldn't it be a `case object` anyway so the equality boils down to `otherEdgeFunction eq this`?

errt (changes requested):

OPAL/ide/src/main/scala/org/opalj/ide/problem/EdgeFunction.scala
Rkoer: I am sorry, but i don't know how this can be done in the presence of generics. Should i use Value=IDEValue and cast the edge function everywhere where the specific type is expected?

OPAL/tac/src/main/scala/org/opalj/tac/fpcf/analyses/ide/instances/lcp_on_fields/problem/LCPOnFieldsEdgeFunctions.scala
Rkoer: It seems like case objects do not override an existing toString implementation of the superclass

OPAL/ide/src/main/scala/org/opalj/ide/problem/EdgeFunction.scala
errt: I think you'd need to make `Value` covariant (i.e., `[+Value <: IDEValue]`) and change the functions taking an `EdgeFunction[Value]` take an `[B >: Value] EdgeFunction[B]` to make it work. In which case I believe it would be `IdentityEdgeFunction extends EdgeFunction[Nothing]` (or `Null`)

OPAL/tac/src/main/scala/org/opalj/tac/fpcf/analyses/ide/instances/lcp_on_fields/problem/LCPOnFieldsEdgeFunctions.scala
errt: Oh, I see. Yes, you can't have a `case object` inheriting from a `case class`

OPAL/tac/src/main/scala/org/opalj/tac/fpcf/analyses/ide/instances/lcp_on_fields/problem/LinearConstantPropagationProblemExtended.scala
```diff
@@ -125,17 +125,10 @@ class LinearConstantPropagationProblemExtended extends LinearConstantPropagation
...
-                case (f: AbstractArrayFact, LCPVariableValue) =>
-                    arrayVar.definedBy.contains(f.definedAtIndex)
-                case _ => false
-            }
-            .map(_._2)
-            .foldLeft(UnknownValue: LinearConstantPropagationValue) {
-                case (value, ArrayValue(initValue, values)) =>
-                    val arrayValue = index match {
+            .collect {
+                case (f: AbstractArrayFact, ArrayValue(initValue, values))
```
errt, line 121: I think it should be possible to merge this into the fold as well

DEVELOPING_OPAL/validate/src/test/java/org/opalj/fpcf/properties/linear_constant_propagation/lcp/ConstantValue.java
```diff
@@ -18,7 +18,7 @@
 @Retention(RetentionPolicy.CLASS)
 public @interface ConstantValue {
     /**
-     * The PC of this variable in the TAC
+     * The PC of this variable in the bytecode
      */
```
errt, line 22: Thanks for updating the documentation, but is it correct now? The code suggests that the values here might not be bytecode PCs but maybe TAC indices.

OPAL/tac/src/main/scala/org/opalj/tac/fpcf/analyses/ide/instances/lcp_on_fields/problem/LinearConstantPropagationProblemExtended.scala
```diff
@@ -144,11 +137,10 @@ class LinearConstantPropagationProblemExtended extends LinearConstantPropagation
                                 VariableValue
                             }
                     }
-                    lattice.meet(value, arrayValue)
-
-                case (_, LCPVariableValue) =>
+                case (f: AbstractArrayFact, LCPVariableValue) if arrayVar.definedBy.contains(f.definedAtIndex) =>
                     VariableValue
             }
+            .foldLeft(UnknownValue: LinearConstantPropagationValue)(lattice.meet)
```
errt: Why is it a `foldLeft` here, but a general `fold` in the other two cases?

OPAL/ide/src/main/resources/reference.conf
```diff
@@ -1,6 +1,4 @@
 org.opalj {
```
errt: Is this config file still required? It is empty now.

OPAL/tac/src/main/scala/org/opalj/tac/fpcf/analyses/ide/instances/lcp_on_fields/problem/LinearConstantPropagationProblemExtended.scala
Rkoer: You mean performing the pattern match inside the fold-callback?

DEVELOPING_OPAL/validate/src/test/java/org/opalj/fpcf/properties/linear_constant_propagation/lcp/ConstantValue.java
Rkoer: I had the same thought, but it originates from `CFG.foreachSuccessor`. Maybe i misunderstood the corr. code?

OPAL/ide/src/main/scala/org/opalj/ide/problem/EdgeFunction.scala
Rkoer: Ah ok, i tried to integrate this.

OPAL/ide/src/main/scala/org/opalj/ide/solver/IDEAnalysis.scala
Rkoer: The most of this (e.g. phase 1, phase 2, sp, n, rs, qs, d5, ...) are named as done in the original algorithm. This helps to associate the paths from the Scala Code with the parts in the Pseudocode of the foundation paper. I'll think about, how this can be made clearer, e.g. with additional documentation.

OPAL/tac/src/main/scala/org/opalj/tac/fpcf/analyses/ide/instances/lcp_on_fields/problem/LinearConstantPropagationProblemExtended.scala
errt: Yes, it seems that should be possible

DEVELOPING_OPAL/validate/src/test/java/org/opalj/fpcf/properties/linear_constant_propagation/lcp/ConstantValue.java
errt: CFG is a generic class agnostic of the underlying code representation. CFGs of bytecode will have PCs, but CFGs of TAC will have TAC indices.

OPAL/ide/src/main/scala/org/opalj/ide/solver/IDEAnalysis.scala
errt: I figured this by now. You are right that keeping close to the original algorithm is a good idea, but documentation would be really helpful here. Both a link to the paper itself and some code documentation that explains what different parts, methods, variables, etc. are for.

OPAL/tac/src/main/scala/org/opalj/tac/fpcf/analyses/ide/instances/lcp_on_fields/problem/LCPOnFieldsProblem.scala
```diff
@@ -180,9 +174,11 @@ class LCPOnFieldsProblem(
                     }
 
                     seenStmts.addAll(workingStmts)
-                    workingStmts = workingStmts.foldLeft(mutable.Set.empty[JavaStatement]) { (nextStmts, stmt) =>
-                        nextStmts.addAll(icfg.getNextStatements(stmt))
-                    }.diff(seenStmts)
+                    workingStmts = workingStmts
+                        .map(icfg.getNextStatements)
```
errt, line 179: Why did you pull this out? It seems the former version was probably more efficient. General rule of thumb: Try not to chain any collection transforming methods (map, filter, fold, etc.) since they create intermediate collections. If it can't be avoided, at least perform them on an iterator to avoid the intermediate collections.

OPAL/tac/src/main/scala/org/opalj/tac/fpcf/analyses/ide/instances/lcp_on_fields/problem/LinearConstantPropagationProblemExtended.scala
Rkoer: I applied this to one occurrence in LCPOnFieldsProblem (see commit 57452f63). IMO this makes the code more complicated. Mainly, because the fold operates on a different data type than the collect (this is what previously was done by the separate map operator). So this requires the fold to operate on type Object instead of the specific types and requires a cast at the end. As fold requires associativeness (compared to foldLeft) and Object is used as type it is also required to catch more cases in the fold (only differing in the tuple order). Please let me know whether i should apply this to the other occurrences too or rather stay with collect and fold combination.

DEVELOPING_OPAL/validate/src/test/java/org/opalj/fpcf/properties/linear_constant_propagation/lcp/ConstantValue.java
Rkoer: In this case i would rename the whole parameter in the JavaStatement class from "pc" to "tacIndex" as this implementation is for TAC. Same for the validation annotations. Is this a suitable solution in your opinion?

errt: Yes, that seems a reasonable solution. PCs tend to be more stable though (we're using a fixed compiler so they will never change, while TAC indices could change if we changed the implementation).

OPAL/tac/src/main/scala/org/opalj/tac/fpcf/analyses/ide/instances/lcp_on_fields/problem/LinearConstantPropagationProblemExtended.scala
errt: It probably only works properly with `foldLeft` there due to the different types. Alternatively, if you want to keep the collect+fold, do it on an iterator. Otherwise, the collect will create a new Set which is both kind of costly (for sets especially, since the uniqueness constraint has to be checked for every entry) and takes up memory, but this is not necessary since we immediately process the values anyway and don't need them stored and also we don't need the set property (i.e., unique elements) since the fold is idempotent for identical elements anyway.

OPAL/tac/src/main/scala/org/opalj/tac/fpcf/analyses/ide/instances/lcp_on_fields/problem/LCPOnFieldsProblem.scala
Rkoer: I did it to replace the foldLeft by a fold. But I can revert this if you want me to. For now i adjusted it to use the iterator solution.

OPAL/tac/src/main/scala/org/opalj/tac/fpcf/analyses/ide/instances/lcp_on_fields/problem/LinearConstantPropagationProblemExtended.scala
Rkoer: Ok, I understand. I modified the code to use the iterator solution.

OPAL/tac/src/main/scala/org/opalj/tac/fpcf/analyses/ide/instances/lcp_on_fields/problem/LCPOnFieldsProblem.scala
errt: I don't see why the foldLeft would be bad. But you can also do it using an iterator. Just avoid the intermediate collections where possible.

OPAL/tac/src/main/scala/org/opalj/tac/fpcf/analyses/ide/solver/JavaForwardICFG.scala
Rkoer: I've put my suggestion for this in PR#300

errt (approved):
LGTM now

Issue #230, opened 2024-11-08
Consider more JDK Types to be Instantiated when Building Call Graphs


johannesduesing:
**Problem**
In #229 we found that calls to `java/io/PrintStream.println` are not found for CG-Algorithms besides CHA **when loading the JDK as interfaces**. The reason is that the type iterator will not consider the type `PrintStream` to be instantiated, as any code instantiating it is not actually loaded.

**Solution** 
There might / could already be some handling for this in place that either does not work anymore or does not cover `PrintStream` - see discussion in #229. Look into the issue and fix it, so that calls to `println` are actually found when loading the JDK as interfaces.

johannesduesing:
I'm currently looking into this issue. For RTA, it's actually an easy fix, as we thought we just need to add some types to the BR `references.conf` under the key `br.analyses.cg.InitialInstantiatedTypesKey.instantiatedTypes`. For now, i added `["java/io/PrintStream", "java/io/InputStream", "java/lang/ClassLoader", "java/lang/Thread"]`, which represent `System.out` and `System.in`, as well as all preconfigured entry points. After publishing this locally, RTA finds calls to `println` for a simple HelloWorld program when loading the `RTJar` as interfaces.

However, for all `PropagationBasedCallGraphKey`s it still does not work. I looked into their implementation i am not sure that they support this form of preconfiguration as it currently stands. The `tac.fpcf.analyses.cg.xta.InstantiatedTypesAnalysis` has a method called [`assignInitialTypeSets`](https://github.com/opalj/opal/blob/f8a8fc179bd99697c1e08076a1fd9c8340cca824/OPAL/tac/src/main/scala/org/opalj/tac/fpcf/analyses/cg/xta/InstantiatedTypesAnalysis.scala#L299) that defines the set of instantiated types for every entrypoint - but it only ever considers types that are passed as parameters to the entrypoint method. Therefore, defining a type as "instantiated no matter the entry point" does not seem to be possible.

I'll try and see what can be done to improve this, just leaving this comment here for documentation purposes.

errt:
Thanks for your investigation!

For RTA: Question is, do we want this. These types are instantiated by the JDK, but on other system, e.g., on Android, they might not be. We could of course decide that this is the default and everyone analyzing a different system should reconfigure their analysis then. At least we should document it though and perhaps provide an AndroidProject.conf that overrides this suitably.

[...]

PR #241, opened 2024-11-29
Improve handling of `java/lang/System` entry points in CG generation

johannesduesing:
This PR resulted from issues discussed in #230, in particular the fact that calls to `System.out.println` are not found by OPAL with multiple CG algorithms when the `RTJar` is loaded (either interfaces-only or fully). The reasons for this are depending on the CG algorithm and configuration, but boil down to:
* RTA cannot find an instantiation of `PrintWriter` (the type of `System.out`) when `RTJar` is loaded as interfaces only - it is simple never instantiated
* Somewhere between JDK9 and JDK11, they switched from using `System.initializeSystemClass` (which we model as an initial entrypoint in the config) to using `System.initPhase1`, `System.initPhase2` and `System.initPhase3` - all of which are called from native code and currently **not** configered to be initial entry points. This is another reason why RTA - even when the RTJar is loaded fully - does not find calls to `println`.
* Propagation-Based algorithms never found a write to `System.out` that initializes the field (even when `RTJar` is loaded fully) - the reason for this is that the fields are set in native code

*Side-Note*: This was never a problem when the `RTJar` was **not** loaded, because `CallGraphAnalysis.unknownLibraryCall` caught any calls to virtual methods (like `println` in this scenario).

**Changes in this PR**
This PR introduces some changes to tackle the problems mentioned above. This includes:
* Declare `initPhase1`, `initPhase2` and `initPhase3` of type `java/lang/System` to be initial entry points in the config - see [the OpenJDK implementation](https://github.com/openjdk/jdk/blob/4da7c3548436ffffb009828891df0d13d47370e3/src/java.base/share/classes/java/lang/System.java#L1753) for details on this.
* Always consider `java/lang/Thread`, `java/lang/ClassLoader`, `java/io/PrintStream` and `java/io/InputStream` to be instantiated via config
* Improve propagation-based cg algorithms - they now always assume that `System.out`, `System.in` and `System.err` are instantiated with their declared types by adding them to the respective `typeSetEntity` - furthermore, we now always assume that the configured instantiated types are available for the `ExternalWorld` type set entity.

With these changes CHA, RTA and all Propagation-Based CG algorithms will now find calls to `System.out.println`, regardless of whether the `RTJar` is loaded at all, as interfaces only, or fully.

**What's still to do**

- [x] Find more types that should always be considered instantiated - if possible define a method for identifying those types and add them to the config
- [x] Analyze why the CFA-Algorithms both do not find calls to `println` and fix the issue
- [x] There is still an issue with library fields if the library is loaded as interfaces only. We will likely have no writes to those fields, as the library method bodys are not loaded, therefore propagation-based algorithms will never consider the fields instantiated. They are also not considered `ExternalWorld` because their definition is available, therefore they cannot be treated differently compared to "normal" project fields. Decide on whether those cases deserve special handling or not - we might just have to live with that fact. The only way of dealing with the would be something like `if(isLibraryField(field) && librariesAreInterfacesOnly) considerInstatiated(typeSetEntity(field), field.fieldType)`


errt:
Thank you for the extensive investigation and the PR. I fully agree with the additions to the config and with adding instantiated types to `ExternalWorld`.

I'm a little wary however with `System.{in,out,err}` special handling for the propagation-based CGs. Could this not lead to scenarios where RTA is less sound than the propagation-based CGs because RTA doesn't have this modeling? Also, different to changes to the config, this can never be overridden for an analysis of a Java System that might behave differently (Android maybe?). Is there another way this could be achieved, e.g., modeling the native methods that set the fields?
If we decide to keep your approach, I'd suggest to make it more general by not hardcoding the `System` class into it. It could also be made configurable via the config files which fields should be initialized.

For the last future work item: That is  a fair point. I think we could deal with it like you suggested, though maybe rather than setting them to the field type, they could access the `ExternalWorld` like they would without loading the interfaces? That would also solve the issue above with `System.{in,out,err}`, right?

johannesduesing:
I think you make a good point. I restructured parts of this PR so that now:
* There is an `InitialInstantiatedFieldsKey` and corresponding `InstantiatedFieldsFinder` to detect fields that shall always be considered instantiated
* Very similar to the `InstantiatedTypesFinder` it has a configurable aspect so that users can specify fields to be considered instantiated. By default the field type will be considered only, but specific types and subtype-trees can be specified in the config.
* The `SoundLibraryInstantiatedFieldsFinder` (implemented but not used by default) would consider all library fields to be instantiated with their declared type if libraries are interfaces only - this could address the last future work item

I think this might be a good solution to the problem - analyses for android can just remove the three lines from the configuration, and they also don't introduce any overhead of libraries are not loaded at all. 

Regarding your suggestion for associating library fields with `ExternalWorld` if libraries are interfaces only: I had that exact solution implemented earlier, but i does not really scale to *any* library besides the `RTJar` - users would have to manually specify the types of their library fields every time they use a new / different library.

What do you think of this idea, does that work for you?

johannesduesing:
Oh i forgot to mention: The `rta.InstantiatedTypesAnalysis` and `xta.InstantiatedTypesAnalysis` now both use the `InitialInstantiatedFieldsKey` for their respective initialization, thus (hopefully) fixing the issue that RTA could sometimes be less sound than XTA.

errt:
This solution sounds reasonable. Thank you for your work!

About the `ExternalWorld`: true for _manual_ specification of types. A suitable `InitialInstantiatedTypesFinder` would do this automatically however.

johannesduesing:
My latest commit adds support for the initial intantiated fields to CFA-1-0 and CFA-1-1. For the latter i had to create dummy allocation sites as the fields are defined in native code - i don't know if that is the way to go, but it seems to work as expected.

So regarding the library fields and the `ExternalWorld` we now have two options:
* Leave it as it currently is. We have the `SoundLibraryInstantiatedFieldsFinder` that can detect those fields and mark them as instantiated. All CG algorithms respect the initial instantiated fields in their respective initialization phases.
* Explicitly associate library fields (if libraries are interfaces only) with the `ExternalWorld` set entity. This will really only work with Propagation-Based algorithms, since RTA and CFA-1-* do not work on set entities (at least i did not see them using `ExternalWorld` anywhere). It would require implementing a suitable `InitialInstantiatedTypesFinder` that would maybe rely on the `InitialInstatiatedFieldsFinder`. The downside would be that we would have to adapt the CG algorithm(s) when it comes to finding instantiated fields in order to encode this specific case.

Is there any specific reason you think we should use the second option? I'm currently thinking the first one should be less overhead, but maybe i'm missing something .. 🤓

errt:
You can have a look at `ConfiguredMethodsPointsToAnalysis`, it needs to do the same for allocation sites in native or otherwise unavailable methods.

The main reason for choosing the second option would be to have fewer necessary modifications. For RTA, it would work immediately, as RTA just uses instantiated types and thus the proper InitialInstantiatedTypesFinder will automatically make it work. For the propagation-based ones it would be just the single line that adds the InitialInstantiatedTypes to the ExternalWorld that we probably want to have anyway. The InstantiatedFieldsFinders would not be necessary. For the points-to-based algorithms, it should be similar: one handling for instantiated types (that we probably need anyway if we don't yet have it) instead of that and the additional one for the fields.
So my question is, where do you see the benefit of the first option? The one that I see is that it could be more specific, in adding types only to individual fields instead of a (potentially large) set of initially instantiated types being available through all (matching - but that could still be many, e.g., `Object` fields) fields.

OPAL/br/src/main/scala/org/opalj/br/analyses/cg/InitialInstantiatedFieldsKey.scala
```diff
@@ -0,0 +1,44 @@
+/* BSD 2-Clause License - see OPAL/LICENSE for details. */
+package org.opalj.br.analyses.cg
```
errt: Please fix package declaration to OPAL's standard of one package per line

errt (changes requested):

johannesduesing:
There are two points why i think the first option may be the better one:
* Like you said, it's more specific and should keep loss of precision to a minimum for Propagation-Based algorithms and CFA - if we load large libraries as interfaces, the `ExternalWorld` could become very large.
* The second option does not work for fields being set in native code if the library contents are actually loaded. If we go with the second option, at least for `System.{out, err, in}` the callgraphs will not capture `println` if the `RTJar` is actually loaded - it works if libraries are interfaces only.

So in my mind, if we want to capture fields being set in native code - and i'm open to discussing whether or not we want that - then we need an `InitialInstatiatedFieldsFinder`, otherwise it won't work in the case that libraries are actually fully loaded.

That being said, if you want me to implement the second option i'll do that - justed wanted to point that out 😄

errt:
Shouldn't the second point already be addressed by the configured native methods analyses? I thought it was.

johannesduesing:
After discussion, we will proceed as follows:
* Remove InitialInstantiatedFieldsFinder
* Use configured native methods to model effects of `setOut0` & others
* Add more instantiated types, see [here](https://github.com/openjdk/jdk/blob/308357cba706478598ef231887e0584eaae839e0/src/hotspot/share/runtime/threads.cpp#L295) 
* Also model effects of `initPhase1` and others so calls can be detected if RTJar is interface only

johannesduesing:
After some further inspection i think this is not going to be as easy as we thought it would. I found that:
* `setOut0` and others are already modeled with their field-setting effects in the config - with the new entrypoints in the config as well, RTA out of the Box finds calls to `System.out.println` **if the RTJar is loaded**
* I was confused that Propagation-Based Algorithms did not find those calls with the RTJar loaded. I found that the `PropagationBasedCallGraphKeys` use the `rta.ConfiguredNativeMethodsInstantiatedTypesAnalysisScheduler` (see [here](https://github.com/opalj/opal/blob/f8a8fc179bd99697c1e08076a1fd9c8340cca824/OPAL/tac/src/main/scala/org/opalj/tac/cg/PropagationBasedCallGraphKeys.scala#L64)). This does not make sense to me, this analysis was specifically written for RTA and attaches all configured instantiated types to the Project entity - as far as my understanding of any of the propagation-based CG algorithms goes, this is conceptually not compatible with XTA & co, which use typeSetEntities to attach types to.
* When loading the `RTJar` as interfaces only, i am facing another issue: The EntryPointsFinder does not consider methods without bodies. This means that even though i carefully configured the effects of `System.initPhase1` with its instantiations and invocations, it is simply never considered by the CG framework, which expects entrypoints to have a body.

If my assessment is correct, i think the following steps would be sensible to take next:
* Write a new `ConfiguredNativeMethodsInstantiatedTypesAnalysis` for XTA & co. I started doing this, but found that because the configuration can state that "This fields is assigned the value of the first parameter" (which is what we have for `setOut0`), it needs to be a lot more complex than the one for RTA, as it would need continuations and dependencies on its caller property - maybe it would be sensible to integrate it into the existing TypePropagationAnalysis for XTA? (It would be structurally very similar and use a lot of the same constructs)
* Consider entrypoints even if they are empty, provided that native method configurations exist for them - unfortunately the correpsonding configuration (plus deserialization utilities) are in the TAC project while entrypoints are in BR, meaning i can't access them directly. I removed the emptiness check for debugging purposes and it did not work as expected out of the box, so this would likely require more debugging

What do you think of this? Is this the right way to go?

errt:
You are right, they are configured to use the RTA class there, which doesn't seem to make much sense. Not sure why that is, I thought there already was a variant for XTA, but apparently, there isn't. So yes, it looks like we need to add that.

For the other issue, it looks like entry points should always be considered, even without a body. I don't think it is necessary that they have configuration attached, they could still just exist without any callees. They should probably _exist_ (so you don't get spurious methods in the CG that are configured as entry point for a different setup than you analyze), but bodies should not be necessary. I wonder why removing the emptiness check didn't just work, so yes, that should be looked into.

Thank you for your investigation!

johannesduesing:
In my opinion, this PR is done now. My local experiments showed that OPAL can now correctly detect calls to `System.out.println` for CHA, RTA, XTA, 1-0-CFA and 1-1-CFA when the JRE is not loaded, loaded as interfaces only or fully loaded. Many things changed throughout the development, so here's the most important points:
* The `EntryPointsFinder` now works on `DeclaredMethod`s instead of `Method`s. Consequently, it now propagates configured entry point methods even if the do not exist in the project - they are then represented as `VirtualDeclaredMethods`.
* All call graph related analyses- especially those related to reading configured behavior - can now deal with the fact that entry points may be virtual
* `DefinitionSites` now use `DeclaredMethod`s instead of `Method`s. This enables us to propagate virtual definition sites for configured points-to behavior in methods that are not actually part of the project. The points-to framework has been adapted to deal with this change.
* The XTA framework got its own `ConfiguredNativeMethodsInstantiatedTypesAnalysis` - previously it used the same as RTA, which did work in the XTA context.

Let me know what you think of those changes - i'll wait for the IT tests to complete and see if the changes broke any test cases.

OPAL/br/src/main/resources/reference.conf
```diff
@@ -97,7 +100,8 @@ org.opalj {
           #analysis = "org.opalj.br.analyses.cg.LibraryInstantiatedTypesFinder"
 
           instantiatedTypes = [
-
+            "java/lang/ClassLoader",
```
errt, line 105: Can you document why these are instantiated?
Also, if Thread is considered instantiated, we may want to consider whether to add ThreadGroup as well.

OPAL/br/src/main/scala/org/opalj/br/analyses/cg/EntryPointFinder.scala
```diff
@@ -57,7 +65,7 @@ trait ApplicationEntryPointsFinder extends EntryPointFinder {
                     && (m.descriptor == MAIN_METHOD_DESCRIPTOR)
                     && (m.name == "main") =>
                 m
-        }
+        }.map(declaredMethods.apply)
```
errt: It would probably be more efficient to move the apply into the the collect, avoiding an additional intermediate collection. Additionally or alternatively, it could be more efficient to perform the collect on an iterator.

OPAL/br/src/main/scala/org/opalj/br/analyses/cg/EntryPointFinder.scala
```diff
@@ -143,7 +153,7 @@ trait LibraryEntryPointsFinder extends EntryPointFinder {
             if (isEntryPoint(method))
                 eps.append(method)
         }
-        super.collectEntryPoints(project) ++ eps
+        super.collectEntryPoints(project) ++ eps.map(declaredMethods.apply)
```
errt: As above, performing the apply immediately instead of with a `map` would remove an intermediate collection. The code above could also probably be improved (along the lines of `val eps = project.allMethodsWithBody.iterator.filter(isEntryPoint)`

OPAL/br/src/main/scala/org/opalj/br/analyses/cg/EntryPointFinder.scala
```diff
@@ -258,32 +271,31 @@ trait ConfigurationEntryPointsFinder extends EntryPointFinder {
...
-                            OPALLogger.warn(
+                        entryPoints = entryPoints ++ methods.map(declaredMethods.apply).toSet
+
+                    case None if !isSubtype =>
+
+                        if (methodDescriptor.isDefined) {
+                            val virtualEntry =
+                                declaredMethods(classType, classType.packageName, classType, name, methodDescriptor.get)
+
+                            entryPoints = entryPoints ++ Set(virtualEntry)
```
errt: The intermediate Set here is probably not necessary

OPAL/br/src/main/scala/org/opalj/br/analyses/cg/EntryPointFinder.scala
```diff
@@ -351,9 +363,11 @@ object AllEntryPointsFinder extends EntryPointFinder {
...
         InitialEntryPointsKey.ConfigKeyPrefix + "AllEntryPointsFinder.projectMethodsOnly"
 
-    override def collectEntryPoints(project: SomeProject): Iterable[Method] = {
+    override def collectEntryPoints(project: SomeProject): Iterable[DeclaredMethod] = {
+        val declaredMethods = project.get(DeclaredMethodsKey)
+
         if (project.config.as[Boolean](ConfigKey))
-            project.allProjectClassFiles.flatMap(_.methodsWithBody)
-        else project.allMethodsWithBody
+            project.allProjectClassFiles.flatMap(_.methodsWithBody).map(declaredMethods.apply)
```
errt: Can probably pull the apply into the flatMap

OPAL/tac/src/main/resources/reference.conf
```diff
@@ -1354,6 +1354,109 @@ org.opalj {
           // java.lang.System                                                                     //
           //                                                                                      //
           //////////////////////////////////////////////////////////////////////////////////////////
+          {
```
errt, line 1357: Please add some (short) documentation on why these methods are included and what they do.

OPAL/tac/src/main/scala/org/opalj/tac/fpcf/analyses/cg/TypeIterator.scala
```diff
@@ -1001,7 +1002,7 @@ class AllocationSitesPointsToTypeIterator(project: SomeProject)
                     result,
                     typesProperty(
                         field,
-                        DefinitionSite(method, defPC),
+                        DefinitionSite(declaredMethods(method), defPC),
```
errt: Remove this apply, use `definedMethod` from line 990 (and probably remove `method` from line 991 completely.

OPAL/tac/src/main/scala/org/opalj/tac/fpcf/analyses/cg/TypeIterator.scala
```diff
@@ -1245,13 +1248,22 @@ class CFA_k_l_TypeIterator(project: SomeProject, val k: Int, val l: Int)
 
                 result = combine(
                     result,
-                    typesProperty(field, DefinitionSite(method, defPC), depender, calleeContext, theTAC.stmts)
+                    typesProperty(
+                        field,
+                        DefinitionSite(declaredMethods(method), defPC),
```
errt: Same as above, this is already available from line 1231, which also means the DeclaredMethodsKey should not be necessary here

OPAL/tac/src/main/scala/org/opalj/tac/fpcf/analyses/cg/xta/InstantiatedTypesAnalysis.scala
```diff
@@ -326,24 +324,30 @@ class InstantiatedTypesAnalysisScheduler(
         // to be pre-initialized. Strings and classes can be introduced via constants anywhere.
         // TODO Only introduce these types to the per-entity type sets where constants are used
         initialize(p, UIDSet(ClassType.String, ClassType.Class))
+        initialize(ExternalWorld, initialInstantiatedTypes)
```
errt, line 325: Not sure why String and Class are initialized for the project, but these other types for the external world.

OPAL/tac/src/main/scala/org/opalj/tac/fpcf/analyses/cg/xta/TypePropagationAnalysis.scala
```diff
@@ -65,21 +66,27 @@ final class TypePropagationAnalysis private[analyses] (
...
+        } else None
 
-        val typeSetEntity = selectTypeSetEntity(definedMethod)
+        val typeSetEntity = selectTypeSetEntity(callContext.method)
         val instantiatedTypesEOptP = propertyStore(typeSetEntity, InstantiatedTypes.key)
-        val calleesEOptP = propertyStore(definedMethod, Callees.key)
-        val readAccessEOptP = propertyStore(method, MethodFieldReadAccessInformation.key)
-        val writeAccessEOptP = propertyStore(method, MethodFieldWriteAccessInformation.key)
+        val calleesEOptP = propertyStore(callContext.method, Callees.key)
+        val readAccessEOptP = methodOpt.map(m => propertyStore(m, MethodFieldReadAccessInformation.key)).orNull
```
errt: Shouldn't we do this with `Option`s instead of `null`s?

OPAL/tac/src/main/scala/org/opalj/tac/fpcf/analyses/cg/xta/TypePropagationAnalysis.scala
```diff
@@ -65,21 +66,27 @@ final class TypePropagationAnalysis private[analyses] (
...
     ): ProperPropertyComputationResult = {
 
-        val definedMethod = callContext.method.asDefinedMethod
-        val method = definedMethod.definedMethod
+        val methodOpt = if (callContext.method.hasSingleDefinedMethod) {
+            Some(callContext.method.definedMethod)
+        } else None
 
-        val typeSetEntity = selectTypeSetEntity(definedMethod)
+        val typeSetEntity = selectTypeSetEntity(callContext.method)
```
errt: Maybe extract callContext.method to a variable since it is used multiple times.

OPAL/tac/src/main/scala/org/opalj/tac/fpcf/analyses/cg/xta/TypePropagationAnalysis.scala
```diff
@@ -373,7 +384,9 @@ final class TypePropagationAnalysis private[analyses] (
 
         // If the call is not static, we need to take the implicit "this" parameter into account.
         if (callee.hasSingleDefinedMethod && !callee.definedMethod.isStatic ||
-            !callee.hasSingleDefinedMethod && !bytecode.instructions(pc).isInvokeStatic
+            // If we can't ask the target if it is static, we look in the bytecode. If we do not have any bytecode,
+            // we soundly assume that the call might be virtual, and we need to add "this".
```
errt: Remove the comma

OPAL/tac/src/main/scala/org/opalj/tac/fpcf/analyses/cg/xta/TypePropagationAnalysis.scala
```diff
@@ -388,15 +401,17 @@ final class TypePropagationAnalysis private[analyses] (
...
+        pc:          Int,
+        bytecodeOpt: Option[Code]
     )(
         implicit
         state:          State,
         partialResults: ArrayBuffer[SomePartialResult]
     ): Unit = {
-        val returnValueIsUsed = {
+        // If the method body is not available, we have to assume the return value might be used
+        val returnValueIsUsed = if (!state.methodHasBody) true
```
errt, line 415: Could think about writing this as `= !state.methodHasBody || {`, but not sure if that is better.

OPAL/tac/src/main/scala/org/opalj/tac/fpcf/analyses/cg/xta/TypePropagationState.scala
```diff
@@ -50,6 +50,9 @@ final class TypePropagationState[ContextType <: Context](
     var methodWritesArrays: Boolean = false
     var methodReadsArrays: Boolean = false
 
+    def methodHasBody: Boolean =
+        callContext.method.hasSingleDefinedMethod && callContext.method.definedMethod.body.isDefined
```
errt, line 54: Should we deal with `MultipleDefinedMethods`?

OPAL/tac/src/main/scala/org/opalj/tac/fpcf/analyses/cg/xta/TypePropagationState.scala
```diff
@@ -114,15 +117,15 @@ final class TypePropagationState[ContextType <: Context](
     var seenIndirectWriteAccesses: IntMap[Int] = IntMap.empty
 
     def readAccessDependee: Option[EOptionP[Method, MethodFieldReadAccessInformation]] = {
-        if (_readAccessDependee.isRefinable) Some(_readAccessDependee) else None
+        if (_readAccessDependee != null && _readAccessDependee.isRefinable) Some(_readAccessDependee) else None
```
errt: Should probably document the potential nullness of these fields

OPAL/tac/src/main/scala/org/opalj/tac/fpcf/analyses/cg/xta/TypePropagationTrace.scala
```diff
@@ -78,16 +77,18 @@ private[xta] class TypePropagationTrace {
             val calleesEOptP = ps(method, Callees.key)
             if (calleesEOptP.hasUBP)
                 calleesEOptP.ub.callSites(typeIterator.newContext(method)).flatMap(_._2)
-            else Iterator.empty
+            else Iterator.empty[Context]
         }
         traceMsg(
-            s"init: ${simplifiedName(method)} (initial types: {${initialTypes.map(simplifiedName).mkString(", ")}}, initial callees: {${initialCallees.map(simplifiedName).mkString(", ")}})"
+            s"init: ${simplifiedName(method)} (initial types: {${initialTypes.map(simplifiedName).mkString(
```
errt: Check if formatting can be improved here

OPAL/tac/src/main/scala/org/opalj/tac/fpcf/analyses/escape/AbstractEscapeAnalysis.scala
```diff
@@ -424,7 +424,8 @@ trait AbstractEscapeAnalysis extends FPCFAnalysis {
     protected[this] def determineEscapeOfDS(
         dsl: (Context, DefinitionSiteLike)
     ): ProperPropertyComputationResult = {
-        val ctx = createContext(dsl, dsl._2.method)
+        // TODO: Handle VirtualDeclaredMethods
```
errt: What about this TODO? Should this `targetMethod` also just be a `DeclaredMethod`?

OPAL/tac/src/main/scala/org/opalj/tac/fpcf/analyses/pointsto/ConfiguredMethodsPointsToAnalysis.scala
```diff
@@ -177,7 +177,8 @@ abstract class ConfiguredMethodsPointsToAnalysis private[analyses] (
         rhs match {
             case md: MethodDescription =>
                 val method =
-                    typeIterator.expandContext(state.callContext, md.method(declaredMethods), pc)
+                    // Configured Calls are made at PC=0, so we need to expand the context accordingly
```
errt, line 180: But should they?

OPAL/tac/src/main/scala/org/opalj/tac/fpcf/analyses/cg/xta/ConfiguredNativeMethodsInstantiatedTypesAnalysis.scala
```diff
@@ -0,0 +1,296 @@
...
+    final val project:               SomeProject,
+    final val typeSetEntitySelector: TypeSetEntitySelector
+) extends ReachableMethodAnalysis {
+
+    private[this] val declaredFields: DeclaredFields = p.get(DeclaredFieldsKey)
+    private[this] val virtualFormalParameters = project.get(VirtualFormalParametersKey)
+
+    private type State = ConfiguredNativeMethodsTypePropagationState[ContextType]
+
+    // TODO remove dependency to classes in pointsto package
```
errt: What about this TODO?

OPAL/tac/src/main/scala/org/opalj/tac/fpcf/analyses/cg/xta/ConfiguredNativeMethodsInstantiatedTypesAnalysis.scala
```diff
@@ -0,0 +1,296 @@
...
+            .filter(_.pointsTo.isDefined)
+            .map { v => (v.method, v.pointsTo.get) }
+            .toMap
+
+    override def processMethodWithoutBody(callContext: ContextType): ProperPropertyComputationResult = {
+        processMethodInternal(callContext)
+    }
+
+    override def processMethod(callContext: ContextType, tacEP: EPS[Method, TACAI]): ProperPropertyComputationResult = {
+        processMethodInternal(callContext)
```
errt: I think we could just call `processMethodWithoutBody` and have the implementation there.

OPAL/tac/src/main/scala/org/opalj/tac/fpcf/analyses/cg/xta/ConfiguredNativeMethodsInstantiatedTypesAnalysis.scala
```diff
@@ -0,0 +1,296 @@
...
+
+/**
+ * Handles the effect of certain (configured native methods) to the set of instantiated types.
+ *
+ * @author Johannes Düsing
+ */
+class ConfiguredNativeMethodsInstantiatedTypesAnalysis private[analyses] (
+    final val project:               SomeProject,
+    final val typeSetEntitySelector: TypeSetEntitySelector
+) extends ReachableMethodAnalysis {
```
errt, line 54: Should the RTA variant of this analysis also be a `ReachableMethodAnalysis`?

OPAL/tac/src/main/scala/org/opalj/tac/fpcf/analyses/cg/xta/ConfiguredNativeMethodsInstantiatedTypesAnalysis.scala
```diff
@@ -0,0 +1,296 @@
...
+    override def processMethodWithoutBody(callContext: ContextType): ProperPropertyComputationResult = {
+        processMethodInternal(callContext)
+    }
+
+    override def processMethod(callContext: ContextType, tacEP: EPS[Method, TACAI]): ProperPropertyComputationResult = {
+        processMethodInternal(callContext)
+    }
+
+    private def processMethodInternal(callContext: ContextType): ProperPropertyComputationResult = {
+        if (!nativeMethodData.contains(callContext.method)) {
```
errt: Maybe extract callContext.method to a variable since it is used multiple times.

OPAL/tac/src/main/scala/org/opalj/tac/fpcf/analyses/cg/xta/ConfiguredNativeMethodsInstantiatedTypesAnalysis.scala
```diff
@@ -0,0 +1,296 @@
...
+            implicit val partialResults: ArrayBuffer[SomePartialResult] = ArrayBuffer.empty[SomePartialResult]
+
+            processParameterAssignments(unseenTypes)(state, partialResults)
+
+            returnResults(partialResults)(state)
+        case _ =>
+            sys.error("received unexpected update")
+    }
+
+    // Taken from rta TypePropagationAnalysis
```
errt: This doesn't make sense. There's no TypePropagationAnalysis for RTA.
Also, instead of duplicating the code, is there a way to reuse it?

OPAL/tac/src/main/scala/org/opalj/tac/fpcf/analyses/cg/xta/ConfiguredNativeMethodsInstantiatedTypesAnalysis.scala
```diff
@@ -0,0 +1,296 @@
...
+
+        implicit val partialResults: ArrayBuffer[SomePartialResult] = ArrayBuffer.empty[SomePartialResult]
+
+        processParameterAssignments(state.ownInstantiatedTypes)
+        processStaticConfigurations
+
+        returnResults(partialResults)
+    }
+
+    private def processStaticConfigurations(implicit
```
errt, line 135: Document this and the following method, they are not trivial.

OPAL/tac/src/main/scala/org/opalj/tac/fpcf/analyses/cg/xta/ConfiguredNativeMethodsInstantiatedTypesAnalysis.scala
```diff
@@ -0,0 +1,296 @@
...
+
+    private type State = ConfiguredNativeMethodsTypePropagationState[ContextType]
+
+    // TODO remove dependency to classes in pointsto package
+    private[this] val nativeMethodData: Map[DeclaredMethod, Array[PointsToRelation]] =
+        ConfiguredMethods
+            .reader
+            .read(p.config, "org.opalj.fpcf.analyses.ConfiguredNativeMethodsAnalysis")
+            .nativeMethods
+            .filter(_.pointsTo.isDefined)
```
errt, line 66: In the RTA variant of this analysis, there is a case that is not present in this analysis and that I think can not be added with this filter being present: https://github.com/opalj/opal/blob/fc0b1f52b3f8812450e73b4af742e8bdaf11d93e/OPAL/tac/src/main/scala/org/opalj/tac/fpcf/analyses/cg/rta/ConfiguredNativeMethodsInstantiatedTypesAnalysis.scala#L98-L108
This code considers the return type of a method that was not configured to be instantiated if we don't have a method body (i.e., native methods). I believe this was done particularly for methods used in reflection implementing functionality such as Class.get[Declared](Method|Field)[s].
Is this really not necessary for the propagation-based analyses?

errt (changes requested):

errt:
As the failed it:test shows, this PR needs re-generation of the `FPCFAnalysesIntegrationTest` reference files. This seems to be because of changes to the entities for the escape analysis? Given the scope of this PR, a deeper look into the files might be sensible to ensure that nothing else was broken.

OPAL/br/src/main/resources/reference.conf
johannesduesing: I now consider all non-static classes for which we have configured entry points to be instantiated and added documentation. I checked with the OpenJDK implementation, those classes are indeed instantiated by the JVM.

OPAL/br/src/main/scala/org/opalj/br/analyses/cg/EntryPointFinder.scala
johannesduesing: Moved apply into the collect and used an iterator - seems to be consensus that this is the best way in terms of performance.

OPAL/br/src/main/scala/org/opalj/br/analyses/cg/EntryPointFinder.scala
johannesduesing: Same as above 👍

OPAL/br/src/main/scala/org/opalj/br/analyses/cg/EntryPointFinder.scala
johannesduesing: Removed unecessary set

OPAL/br/src/main/scala/org/opalj/br/analyses/cg/EntryPointFinder.scala
johannesduesing: Did that

OPAL/tac/src/main/resources/reference.conf
johannesduesing: Added documentation on why methods are included and brief description on each individual points-to assignment.

OPAL/tac/src/main/scala/org/opalj/tac/fpcf/analyses/cg/xta/InstantiatedTypesAnalysis.scala
johannesduesing: I think i inherited this code, it says in the comment above that initializing `String` and `Class` may be required for *"Some cooperative analyses originally meant for RTA"* - i do not know if this is still the case, without the `initialize(p, [...])` the code still works on my testbench...

OPAL/tac/src/main/scala/org/opalj/tac/fpcf/analyses/cg/xta/TypePropagationAnalysis.scala
johannesduesing: Did that

OPAL/tac/src/main/scala/org/opalj/tac/fpcf/analyses/cg/xta/TypePropagationAnalysis.scala
johannesduesing: Which one? 😅  If you are referring to the one after "virtual" - the IntelliJ grammar checker tells me to keep it whenver i remove that.

OPAL/tac/src/main/scala/org/opalj/tac/fpcf/analyses/cg/xta/TypePropagationAnalysis.scala
johannesduesing: I honestly think it is more readable the way it is now

OPAL/tac/src/main/scala/org/opalj/tac/fpcf/analyses/cg/xta/InstantiatedTypesAnalysis.scala
errt: The question was not whether we want String and Class here - we do. The question is why are different initially instantiated types here in different sets. It seems that either they should all be in `p` or all in `ExternalWorld`.

OPAL/tac/src/main/scala/org/opalj/tac/fpcf/analyses/cg/xta/TypePropagationState.scala
johannesduesing: Should not be necessary, i made these types an `Option` now

OPAL/tac/src/main/scala/org/opalj/tac/fpcf/analyses/cg/xta/TypePropagationState.scala
johannesduesing: Very good question - the entire `TypePropagationAnalysis` currently does not really support that. It requires either one or no actual implementations. If you think that's necessary i will have a look, but it will require some larger refactoring work.

OPAL/tac/src/main/scala/org/opalj/tac/fpcf/analyses/cg/xta/TypePropagationTrace.scala
johannesduesing: Tried something different, is that better?

OPAL/tac/src/main/scala/org/opalj/tac/fpcf/analyses/cg/xta/TypePropagationAnalysis.scala
errt: Yes. I'm not sure the grammar checker is correct there, since that makes it a new, unrelated sentence as far as I understand. Probably remove the "we" too and it won't complain, though.

OPAL/tac/src/main/scala/org/opalj/tac/fpcf/analyses/cg/xta/TypePropagationState.scala
errt: Let's leave it the way it is then

OPAL/tac/src/main/scala/org/opalj/tac/fpcf/analyses/pointsto/ConfiguredMethodsPointsToAnalysis.scala
johannesduesing: Are you asking whether configured calls should be made at PC=0? If so, i am unsure... We have multiple different analyses that handle the integration of configured values into the call graph. For example for the points-to configuration, the corresponding analysis "makes up" (negative) PC values to assign points-to information to. The analysis responsible for configured method invocations decides to use PC=0 for all calls - since the two of them are not really "communicating" it would be hard to make them agree on one specific "virtual PC layout".

errt: Yes, that was my question. I see the issue. Does this cause any potential problems (especially, if there are multiple invocations configured, could we end up with a situation where the points-to sets end up containing the wrong values)? If not, let's keep it.

OPAL/tac/src/main/scala/org/opalj/tac/fpcf/analyses/cg/xta/ConfiguredNativeMethodsInstantiatedTypesAnalysis.scala
johannesduesing: I'd like to just remove the `TODO` comment. We do not depend on anything from the actual `pointsto` package, only the `tac.fpcf.analyses.PointsToRelation` class which represents the configured data. This is necessary to derive instantiated types information from - i do not see how that would be a problem.

errt: I tend to agree. Though we might consider renaming the class since what it models is rather an assignment than specifically points-to information.

OPAL/tac/src/main/scala/org/opalj/tac/fpcf/analyses/cg/xta/ConfiguredNativeMethodsInstantiatedTypesAnalysis.scala
johannesduesing: Makes sense to me, i changed it to also use the trait 👍

OPAL/tac/src/main/scala/org/opalj/tac/fpcf/analyses/cg/xta/ConfiguredNativeMethodsInstantiatedTypesAnalysis.scala
johannesduesing: I think that was a typo, i meant to say i took it from the xta analysis - in any case, i moved that piece of code to the appropirate package object where i think it makes the most sense.

OPAL/tac/src/main/scala/org/opalj/tac/fpcf/analyses/cg/xta/ConfiguredNativeMethodsInstantiatedTypesAnalysis.scala
johannesduesing: I renamed `PointsToRelation` to `EntityAssignment`, as i think that's essentially what the class is - hope that works for you.

OPAL/tac/src/main/scala/org/opalj/tac/fpcf/analyses/escape/AbstractEscapeAnalysis.scala
johannesduesing: Made the Escape Analysis framework used DeclaredMethods alltogether

OPAL/tac/src/main/scala/org/opalj/tac/fpcf/analyses/cg/xta/InstantiatedTypesAnalysis.scala
johannesduesing: As we figured out together yesterday, i removed the initialization of types for the Project `p` as it should no longer be required.

OPAL/tac/src/main/scala/org/opalj/tac/fpcf/analyses/cg/xta/ConfiguredNativeMethodsInstantiatedTypesAnalysis.scala
```diff
@@ -203,12 +203,17 @@ class ConfiguredNativeMethodsInstantiatedTypesAnalysis private[analyses] (
...
                 val theField = declaredFields(ClassType(cf), name, FieldType(fieldType))
                 val fieldSetEntity = typeSetEntitySelector(theField)
+
+                // Obtain the virtual formal parameter. This will be null if the current method is virtual!
                 val theParameter = pd.fp(state.callContext.method, virtualFormalParameters)
 
                 // Obtain the parameter type
-                val theParameterType = if (theParameter.origin == -1) {
+                val theParameterType = if (theParameter == null || theParameter.origin == -1) {
+                    // If the virtual formal parameter is null (we are in a virtual method) or the origin is unknown,
```
errt, line 239: We might have to have a look into the VirtualFormalParameters(Key) and check whether it should not, like the DeclaredMethodsKey, support creating new VirtualFormalParameters for VirtualDeclaredMethods when queried. The way it is now, it creates them for all DeclaredMethods that are available when the VirtualFormalParametersKey is first computed, which is not deterministic since the DeclaredMethods can be extended any time by querying for a VirtualDeclaredMethod

OPAL/tac/src/main/scala/org/opalj/tac/fpcf/analyses/cg/xta/ConfiguredNativeMethodsInstantiatedTypesAnalysis.scala
johannesduesing: After our discussion yesterday i added this feature to XTA as well

OPAL/tac/src/main/scala/org/opalj/tac/fpcf/analyses/cg/xta/ConfiguredNativeMethodsInstantiatedTypesAnalysis.scala
johannesduesing: I do agree, i thought the same thing when looking into the issue. However, i do not think that should be part of this PR, as i would expect the impact to be relatively big. If that's okay with you, i'll leave this as it is and create an issue for updating the VirtualFormalParameter computations in a separate PR.

errt: Sure, just let's not forget about it.

johannesduesing: I created #316 to document this

johannesduesing:
@errt I updated most expected values for the IT test, i unfortunately do not have the resources at hand to generate the `ExtensiveAnalysesTest-OPAL-MultiJar-SNAPSHOT-01-04-2018-dependencies.txt.gz` file. I looked through the changes of two archives in depth and found that they really only differed in the `toString` representation of methods. If you want me to i can do some further investigation, but i am confident that there are no unintended behavioral changes.

Can you regenerate the missing expected value befor i rerun the IT tests?

errt (approved):
Looks complete now

PR #243, opened 2024-12-03
Fix isIdle in PKESequentialPropertyStore

errt:
The previous implementation forced registering eager computations after lazy ones as they fill the task queue. This implementation, which matches the one in PKECPropertyStore, does not have that limitation.


johannesduesing (approved):
LGTM 👍

Issue #223, opened 2024-10-18
Add requiredProjectInformationKeys to EntryPointFinder


JuliusNmn:
Instead of a doc comment on this, we should extend EntryPointFinder to have a method for requiredProjectInformationKeys, override this here to specify the AndroidManifestKey and then use that method in the InitialEntryPointKey to compute the correct set of required keys. That will need some rewriting of the InitialEntryPoints key to get the used EntryPointsFinder earlier though.

_Originally posted by @errt in https://github.com/opalj/opal/pull/222#discussion_r1806299022_

errt:
Fixed by #244

PR #244, opened 2024-12-05
#223 add requirements method to EntryPointFinder trait

JuliusNmn:
Add requirements method to `EntryPointsFinder` trait to allow specifying required PIK for an `EntryPointFinder`, including checks. #223

### DIFF
--- a/OPAL/br/src/main/scala/org/opalj/br/analyses/cg/EntryPointFinder.scala
+++ b/OPAL/br/src/main/scala/org/opalj/br/analyses/cg/EntryPointFinder.scala
@@ -28,6 +28,13 @@ sealed trait EntryPointFinder {
      * This method must be implemented by any subtype.
      */
     def collectEntryPoints(project: SomeProject): Iterable[Method] = Set.empty[Method]
+
+    /**
+     * Returns ProjectInformationKeys required by this EntryPointFinder
+     * If no extra keys are required, `Nil` can be returned.
+     */
+    /*ABSTRACT*/
+    def requirements(project: SomeProject): ProjectInformationKeys = { Nil }
 }
 
 /**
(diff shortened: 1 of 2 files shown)


OPAL/br/src/main/scala/org/opalj/br/analyses/cg/EntryPointFinder.scala
```diff
@@ -28,6 +28,13 @@ sealed trait EntryPointFinder {
      * This method must be implemented by any subtype.
      */
     def collectEntryPoints(project: SomeProject): Iterable[Method] = Set.empty[Method]
+
+    /**
+     * Returns ProjectInformationKeys required by this EntryPointFinder
+     * If no extra keys are required, `Nil` can be returned.
+     */
+    /*ABSTRACT*/
+    def requirements(project: SomeProject): ProjectInformationKeys = { Nil }
```
errt: Remove the braces. Also, remove the "ABSTRACT" comment. This method is not abstract, it is validly implemented and most entry point finders will not need to override it.

OPAL/br/src/main/scala/org/opalj/br/analyses/cg/InitialEntryPointsKey.scala
```diff
@@ -64,6 +64,11 @@ object InitialEntryPointsKey extends ProjectInformationKey[Iterable[Method], Not
 
         val fqn = entryPointFinder.get
         val epFinder = instantiateEntryPointFinder(fqn)
+
+        for (requiredProjectInformationKey <- epFinder.requirements(project)) {
```
errt: This should not be needed, entry point finders that need a key can (and will obviously) get it themselves. However, what is required is that the InitialEntryPointKey's `requirements` method returns the entry point finder's requirements. Have a look at the CallGraphKey, it works pretty much the same

errt (changes requested):

errt (approved):
LGTM

PR #249, opened 2025-01-20
Fixed some missing uses

19Vik19:
Fixed some missing uses in the DoPrivilegedCGAnalysis and in the ReflectionRelatedFieldAccessesAnalysis.

These missing uses could lead to a non-functioning scheduling order, resulting in an incorrect analysis.


OPAL/tac/src/main/scala/org/opalj/tac/fpcf/analyses/cg/DoPrivilegedCGAnalysis.scala
```diff
@@ -305,7 +305,10 @@ object DoPrivilegedAnalysisScheduler extends BasicFPCFEagerAnalysisScheduler {
     override def requiredProjectInformation: ProjectInformationKeys =
         Seq(DeclaredMethodsKey, VirtualFormalParametersKey, DefinitionSitesKey, TypeIteratorKey)
 
-    override def uses: Set[PropertyBounds] = Set.empty
+    override def uses: Set[PropertyBounds] = PropertyBounds.ubs(Callers, Callees)
+
+    override def uses(p: SomeProject, ps: PropertyStore): Set[PropertyBounds] =
+        p.get(TypeIteratorKey).usedPropertyKinds
```
errt, line 311: I suppose there are more analyses that need that fix. Will have a look and add them to this PR as well.

errt (approved):

PR #250, opened 2025-01-20
Fixed the CallGraph error where the call graph is not calculated beca…

19Vik19:
…use the analysis was not calculated in the first phase of the scheduling process.

When implementing an automatic scheduler, an error was detected that the call graph is not calculated because the call graph should be calculated in a later phase.

When the code was analyzed with the debugger, I discovered that the ReachableMethodAnalysis in the analyze method is aborted by a FinalP(NoCallers).

By changing the register method and deleting the init method, which is normally executed before scheduling, the call graph can also be calculated in later phases.


errt:
This is necessary as a workaround for #112 . If that was fixed, this should have worked. Still, there should be no harm in setting the entry points only when the analysis is actually registered, instead of the global initialization phase before.

errt (approved):

PR #251, opened 2025-01-20
Fixed exception when scheduling the purity analysis with all analyses in one phase

19Vik19:
By scheduling the PurityAnalysis with all analyses in one phase, an exception is thrown in the FieldLocalityAnalysis if new callers are processed by the handleNewCallers method.

### DIFF
--- a/OPAL/tac/src/main/scala/org/opalj/tac/fpcf/analyses/FieldLocalityAnalysis.scala
+++ b/OPAL/tac/src/main/scala/org/opalj/tac/fpcf/analyses/FieldLocalityAnalysis.scala
@@ -714,7 +714,7 @@ class FieldLocalityAnalysis private[analyses] (
         state.addCallersDependee(newEP)
 
         val tacDependee = state.getTACDependee(definedMethod.definedMethod)
-        val oldCallers = oldEP.map(ep => if (ep.hasUBP) ep.ub else null).get
+        val oldCallers = if (oldEP != null && oldEP.get.hasUBP) oldEP.get.ub else null
 
         if (newEP.hasUBP && tacDependee.isDefined && tacDependee.get.hasUBP && tacDependee.get.ub.tac.isDefined) {
             val callers = newEP.ub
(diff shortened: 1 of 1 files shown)


OPAL/tac/src/main/scala/org/opalj/tac/fpcf/analyses/FieldLocalityAnalysis.scala
```diff
@@ -714,7 +714,7 @@ class FieldLocalityAnalysis private[analyses] (
         state.addCallersDependee(newEP)
 
         val tacDependee = state.getTACDependee(definedMethod.definedMethod)
-        val oldCallers = oldEP.map(_.ub).orNull
+        val oldCallers = oldEP.map(ep => if (ep.hasUBP) ep.ub else null).get
```
errt: Thinking about this, it would probably be sensible to do the `get` first, getting rid of the `map` operation that needlessly creates a new `Option` object. The added check whether there actually is an upper bound or else use `null` of course still is necessary.

errt (changes requested):

OPAL/tac/src/main/scala/org/opalj/tac/fpcf/analyses/FieldLocalityAnalysis.scala
19Vik19: Good idea, I added an additional check to avoid a NullPointerException. I hope that change meets your expectations.

errt: No null check needed, `oldEP` comes from a `Map.get`, which never returns null (that's why it's an Option). Also, should avoid the double `get` on the Option. Do it once, then check for the `hasUBP`

19Vik19: Null-check and double-get for the option (oldEP) removed.

errt: Nice solution!

errt (approved):

PR #252, opened 2025-01-21
Fix triggered analyses in PKECPropertyStore

### LINKED ISSUE #112: Triggered analyses not executed when running multiple computations

errt:
This should fix #112


johannesduesing (approved):
LGTM 👍

PR #264, opened 2025-05-07
Fix typo in ScalaDoc

errt:


johannesduesing (approved):
LGTM (went into a little deep dive on whether or not it should technically be `"descriptor of the methods returned"` instead of `"returned methods"`, turns out both options are correct) 😄

PR #265, opened 2025-05-07
Fix assertion on call-graph integrity

errt:
If the project is inconsistent (e.g., a class file has been compiled against a different version of an interface), the call-graph may resolve to abstract methods in a best-effort manor. This is to be expected and thus must not be ruled out by the assertion.


OPAL/tac/src/main/scala/org/opalj/tac/fpcf/analyses/cg/xta/TypePropagationAnalysis.scala
```diff
@@ -268,7 +268,9 @@ final class TypePropagationAnalysis private[analyses] (
             // Instances of DefinedMethod we see should only be those where the method is defined in the class file of
             // the declaring class type (i.e., it is not a DefinedMethod instance of some inherited method).
```
johannesduesing: Shouldn't we technically also adapt this comment, as it is no longer true in the exact scenario you created this PR for?

errt: I merged the other comment into this one to hopefully improve the clarity, let me know what you think.

johannesduesing (approved):
PR looks good to me now 👍

PR #266, opened 2025-05-08
Fix FieldAccessInformationAnalysisDemo

### LINKED ISSUE #236: ClassExceptionError Demo 48
When trying to execute Demo 48 I get the error at the end of this log file:
[log](https://github.com/user-attachments/files/17720533/sbt.output.log)

I get some errors with the other demos, what is the best practice to handle the errors?

errt:
Fixes #236 by using DeclaredFields everywhere


johannesduesing (approved):
I was able to run this demo and the output looks good to me. The changes to `DeclaredFields` could use some ScalaDoc, but currently the class isn't documented at all...so maybe it's not that important at the moment 🤓

PR #268, opened 2025-05-09
Update sonatype release process

errt:
Sonatype's OSSRH service shuts down on June 30th, requiring us to change our release process to the newer Sonatype Central service: https://central.sonatype.org/faq/what-is-different-between-central-portal-and-legacy-ossrh/

More information on https://github.com/xerial/sbt-sonatype

With the next release, we have to check whether more configuration, e.g., POM metadata, tokens, etc. need to be set up.

I took the chance to also include @johannesduesing as a developer in our POM metadata.

### DIFF
--- a/LICENSE
+++ b/LICENSE
@@ -1,5 +1,5 @@
 BSD 2-Clause License:
-Copyright (c) 2009 - 2024
+Copyright (c) 2009 - 2025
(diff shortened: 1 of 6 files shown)


johannesduesing (approved):
Looks good to me, and thanks for the credits!

PR #269, opened 2025-05-09
Fix compatibility issues in Dependency Graph Generation

### LINKED ISSUE #267: Problems with Project Dependency Graph Generation
There are two issues with the SBT command that generates the OPAL sub-project dependency graph (`sbt runProjectDependencyGeneration`):
* It does not work on windows because it invokes `id -u` and `id -g`, which are not (always) defined on windows
* It pulls the latest version of a diagram creation tool (mermaid), which seems to have changed some default behavior. As a result, the diagram no longer contains piecewise-linear arrows (as intended), but curvy arrows instead. Also it no longer renders the `\n` inside graph nodes correctly, but instread just prints the literal characters to the graph node.

The second problem can temporarily be fixed by using version 10.7.0 of mermaid (`minlag/mermaid-cli:10.7.0 `), which renders the graph as we expect it. However, it would be better to find out how we have to configure the latest version of mermaid to produce the graph we want. For the first issue, we should find out if it matters at all which user creates the resulting `.svg` and `.pdf` files - if not, we can remove the usage of `id -u` and `id -g`.

johannesduesing:
This PR addresses #267. The following changes have been applied:

* Added error handling to the task definition of `runProjectDependencyGeneration` so that it will try to obtain the user's id and pass it to the docker container (required on linux), but will not pass any argument if it fails to obtain the id (works for windows). I tested this on Windows and Linux and found no issues, but i'd be happy to have some more feedback on this.
* Replaced `\n` with `<br>` as this seems to be the way to go now for adding line breaks to nodes in mermaid

There is still one issue, which is that arrows are not linear anymore, but curved. This is actually [a known issue in mermaid](https://github.com/mermaid-js/mermaid/releases/tag/mermaid%4011.6.0) and was fixed two months ago. However, the `mermaid-cli` image we are using has not been updated since last year, and does therefore not contain the fix for applying curve styles at the moment. Since we always pull the latest version of `mermaid-cli`, this issue should fix itself once the maintainers upgrade their dependency to mermaid and release a new version. Until then, we could either live with curved lines or use an older version of mermaid (e.g. `10.7.0`) where cruve styles work as intended. Let me know what you think!

Closes #267 .

### DIFF
--- a/OPAL/ProjectDependencies.mmd
+++ b/OPAL/ProjectDependencies.mmd
@@ -1,31 +1,30 @@
 %%{ init: { 'flowchart': { 'curve': 'linear' } } }%%
 flowchart BT
-    
-    common[Common\n common]
+    common[Common<br>common]
 
-    si[Static Analysis Infrastructure\n si]
-    bi[Bytecode Infrastructure\n bi]
+    si[Static Analysis Infrastructure<br>si]
+    bi[Bytecode Infrastructure<br>bi]
 
-    br[Bytecode Representation\n br]
-    da[Bytecode Disassembler\n da]
+    br[Bytecode Representation<br>br]
+    da[Bytecode Disassembler<br>da]
 
-    ifds[IFDS\n ifds]
-    ai[Abstract Interpretation Framework\n ai]
-    bc[Bytecode Creator\n bc]
+    ifds[IFDS<br>ifds]
+    ai[Abstract Interpretation Framework<br>ai]
+    bc[Bytecode Creator<br>bc]
 
-    tac[Three Address Code\n tac]
-    de[Dependency Extraction\n de]
-    ba[Bytecode Assembler\n ba]
+    tac[Three Address Code<br>tac]
+    de[Dependency Extraction<br>de]
+    ba[Bytecode Assembler<br>ba]
 
-    ll[LLVM\n ll]
-    apk[APK\n apk]
-    av[Architecture Validation\n av]
+    ll[LLVM<br>ll]
+    apk[APK<br>apk]
+    av[Architecture Validation<br>av]
 
     framework[Framework]
 
     demos[Demos]
-    bp[BugPicker\n bp]
-    hermes[Hermes\n hermes]
+    bp[BugPicker<br>bp]
+    hermes[Hermes<br>hermes]
 
     style common fill:#9cbecc,color:black
     style framework fill:#c0ffc0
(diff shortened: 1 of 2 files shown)


build.sbt
```diff
@@ -527,10 +527,23 @@ lazy val runProjectDependencyGeneration =  ThisBuild / taskKey[Unit] ("Regenerat
...
+      val uid = "id -u".!!.stripSuffix("\n").toString
+      val gid = "id -g".!!.stripSuffix("\n").toString
+      s"$uid:$gid"
+  } match {
+    case Success(userSpec) =>
+      s"-u $userSpec"
+    case Failure(ex) =>
+        s.log.warn(s"Cannot obtain user id, using mermaid default user for diagram file creation")
+      ""
+  }
```
errt: Wouldn't it be easier to do
```
val dockerUserArg = Try {
      val uid = "id -u".!!.stripSuffix("\n").toString
      val gid = "id -g".!!.stripSuffix("\n").toString
      s"-u $uid:$gid"
  }.getOrElse("")
```
? I don't see a need for the warning on Windows and on Linux, if the Try fails, the user will get a docker error anyway.

johannesduesing: Makes sense, i updated the code. For my testing i also put `id.exe` (shipped with git) on my windows `PATH`, which did not seem to work at the time, but now suddenly does. It returns some weird spaces after the user ids, that's why i added a `.trim()` to both outputs as well. It should now work on Linux, as well as on Windows (both with and without `id.exe` on PATH)

errt (approved):
Looks good to me. We could have a look into whether we can reduce the size of the boxes again, but that is not of importance.

OPAL/ProjectDependencies.mmd
```diff
@@ -1,31 +1,30 @@
 %%{ init: { 'flowchart': { 'curve': 'linear' } } }%%
```
errt, line 1: Should we remove the linear style here since it doesn't seem to be working anymore?

johannesduesing: Well it is the correct way to specify linear curves, and it only [because of a bug](https://github.com/mermaid-js/mermaid/releases/tag/mermaid%4011.6.0) that it is not working with the current latest version of `mermaid-cli`. I think we should leave it in and hope that the bugfix is incorporated into `mermaid-cli` soon, then the style specification should work again.

PR #271, opened 2025-05-13
Bump bytecode support to Java 24

errt:
"Raider, äh, ObjectType heißt jetzt ClassType, sonst ändert sich nix": https://redirect.github.com/YujiSoftware/jvm_spec_diff/compare/java22...java23?w=1#diff-d0a1ed5e273c6656f2d9da4f2ae9898348f2632f5fde28608b96d2d1a11cac0f


TorunR (dismissed):

maximilianruesch (changes requested):
The PR says we upgrade to Java 24, the code agrees but says its from Java 21, the linked diff however is only comparing Java 22 to Java 23.

There are a few changes in the full 21-24 diff of the JVM spec which have caught my interest:
1. There is the renaming of object types to class types which this PR largely consists of
2. There is a weird wording change from MethodHandles to "objects" in https://github.com/YujiSoftware/jvm_spec_diff/compare/YujiSoftware:e4bcdba...YujiSoftware:817facb#diff-bbbf09def7b3db97bb96c044a88c026fd50b81b6b0cc99619947dc36e3928d0eL723-R699 which I could quite get my mind wrapped around
3. There is a wording change from "actual" to "storage" types (in the context opposed to computational types): https://github.com/YujiSoftware/jvm_spec_diff/compare/YujiSoftware:e4bcdba...YujiSoftware:817facb#diff-6e7dfb60c94359b4b8cf24acf973c5923988fe115fce3dcc57573f0976c8455bL544-R546
All other changes should be fine.

As mentioned, 1. is dealt with by this PR (although we do not necessarily have to follow suit with what the JVM calls things, its convenient for future users). I did not find any reference to code mentioning 3. (which would be in similar vein to 1.), but I am unsure if 2. impacts us. I am even unsure how we plan to support multiple JVM versions if they ever conflict with one another (although I suppose breaking backwards compatibility defeats the purpose of Java a bit).

What are your thoughts?

errt:
@maximilianruesch Thanks for your review.
I don't think we're using the term "actual type" in OPAL anywhere (at least not in the sense of the JVM specification).
I am also unsure about number 2, but it might just clarify that bootstrap method arguments are not required to be MethodHandles, but can also be constants, which we already model in OPAL (see the heirs of `BootstrapArgument`), which is why I think this change doesn't affect as.
I found no further changes affecting us in the diffs for 21->22 or 23->24, which is why I linked only the one above.

maximilianruesch (dismissed):
LGTM, after merge conflicts and tests are resolved!

maximilianruesch (approved):

PR #272, opened 2025-05-14
New sbt alias "format" for formatting all files

errt:
Adds a new alias `sbt format` that formats all files (note that `sbt scalafmt` does not format test and integration test files).


johannesduesing (approved):
Works for me 👍

PR #273, opened 2025-05-16
Fix two issues in invokedynamic rewriting

errt:
- There are no append methods on StringBuilder for shorts and bytes, they are appended as ints
- Invalid class names (starting with a `/`) would have been generated if class was in the unnamed package


johannesduesing (approved):
Looks good to me 👍

PR #274, opened 2025-05-17
Repeatable property matcher

Rkoer:
Add AbstractRepeatablePropertyMatcher as proposed in PR#240 about IDE and apply it to the DirectCallMatcher


errt (approved):
LGTM

PR #279, opened 2025-06-06
Add Hermes query for reflective field writes

errt:
Had this still lying around here. No harm in merging it.


johannesduesing (approved):
LGTM 👍

PR #281, opened 2025-06-10
Regenerate FPCFAnalysesIntegrationTest comparison files for renaming ObjectType->ClassType

errt:
Should have seen that coming. As far as I could see, this has not changed any results, just the string representation changed, requiring the regeneration.


johannesduesing (approved):
LGTM, i started it tests just to verify.

PR #277, opened 2025-06-05
Move non-Java-specific code out of br subproject

errt:
Several classes related to projects and the FPCF subsystem are not Java(-bytecode)-specific. This PR moves them to the si subproject to facilitate future analyses of non-Java or cross-language projects.

This implements the ideas of #86, superseding it.

PR #285, opened 2025-06-24
Move non-Java-specific code out of br subproject (cont'd)

errt:
Follow up on #277 
- Fixes an bug introduced in that PR
- Moves an additional class to SI
- Stores the `Project` twice in the `PropertyStore` context (just to be on the safe side)
- Further cleanup


johannesduesing (approved):
Looks good to me, integration tests seem to agree 😄

PR #286, opened 2025-06-24
Adapt tutorials to changes in OPAL

errt:
Tutorials had diverged from OPAL because of the recent move of FPCF classes to SI (#277, #285) and older changes to the call-graph architecture (use of ContextProviders).


johannesduesing (approved):
All changes seems sensible to me 👍

PR #287, opened 2025-07-01
Add framework for data-flow analysis based on structural analysis

errt:
Adds a generic framework for data-flow analysis that uses structural analysis, i.e., the recovery of the structure of nested control structures.

This is a spin-off from #1.


johannesduesing (approved):
This looks good to me, and should be ready to merge. I was thinking (in the spirit of providing documentation and samples) that it might be nice to have a demo on how to use this specific type of analysis, but maybe that's a task for later 🤓

errt:
#1 does use it, but it's not exactly a small demo

PR #288, opened 2025-07-01
Introduce PDUWebs

errt:
PDUWebs are a persistable (i.e., bytecode program counter based) representation of def-use webs (i.e., the union of def-use chains that share common definitions or uses).

This is a spin-off from #1.


johannesduesing (approved):
LGMT 👍

PR #290, opened 2025-07-09
Format check command alias and git hook

errt:
Adds a command alias `sbt checkFormat` similar to `sbt format` (see #272) which does not format files, but instead returns an error code if files need to be formatted. The main use case for this is for use in a git hook.

An example git hook is also included in the PR. If one wants to use it, the file has to be copied to `.git/hooks` (and potentially `chmod +x`). This will then check the formatting, and if necessary, format the files and abort the commit for the developer to review the formatting changes before retrying the commit. Note that due to the need to fire up sbt and scan many files, this significantly increases commit times, but at the same time, it prevents CI failures for incorrectly formatted files. If (local) commits need to be done without proper formatting applied, git commit can be used with `--no-verify` to skip the hook.

### DIFF
--- a/build.sbt
+++ b/build.sbt
@@ -95,6 +95,7 @@ addCommandAlias(
 addCommandAlias("cleanBuild", "; project OPAL ; cleanAll ; buildAll ")
 
 addCommandAlias("format", "; scalafmt; Test / scalafmt; IntegrationTest / scalafmt")
+addCommandAlias("checkFormat", "; scalafmtCheck; Test / scalafmtCheck; IntegrationTest / scalafmtCheck")
 
 lazy val IntegrationTest = config("it") extend Test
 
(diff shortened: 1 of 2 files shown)


johannesduesing (approved):
As discussed, i had a brief look into alternative approaches. It might be possible to define a custom task key in the `build.sbt` file that would do essentially the same thing the `pre-commit` is currently doing - that would require just one instance of SBT to be spawned and might make the hook invocation a bit faster. However, it would come at the cost of defining a new task key that is not really closely related to OPALs core functionality, so i do not think it's the way to go for now. Maybe we'll find a better solution in the future, for now this looks good as it is 👍

errt:
If it's still simple, we could try it. But if it still requires to run both scalafmtCheck and scalafmt, it is probably not that much faster.

johannesduesing:
It does require to run both, i checked the scalafmt docs and did not find a better way - so i guess we're fine leaving it as it is 👍

PR #291, opened 2025-07-16
Fix FieldAssignability re-entrancy issue

errt:
As the same method can be processed multiple times (e.g., when Callers are updated), checking for whether the level is already a lazy intitialization one leads to non-deterministic behaviour. It should also not be necessary, given that the first thing the `isAssignable` method does is to check whether there are multiple non-initializer writes.

Also included some readability improvements by matching tuple elements to descriptive variable names.


johannesduesing (approved):
LGTM 👍

PR #292, opened 2025-07-18
Move engineered bytecode into individual packages

errt:
This avoids having different classes of the same FQN in OPAL's codebase which can lead to non-deterministic results when analyzing OPAL itself based on which class is read first.


johannesduesing (approved):
This seems reasonable, as discussed in person 👍

PR #293, opened 2025-07-21
Fix IgnoreSynchronization domain

errt:
Missed a case where there would be no exception thrown by the monitor statement, resulting in an OperationNotSupportedException.


johannesduesing (approved):
LGTM 👍

PR #294, opened 2025-07-21
Dependency extractor: Add new attributes

errt:
The DependencyExtractor missed the "recently" introduced bytecode attributes: Modules, Records, Nests, permitted subclasses


OPAL/br/src/main/scala/org/opalj/br/VirtualSourceElement.scala
```diff
@@ -66,6 +66,25 @@ object VirtualSourceElement {
...
+     */
+    override def compare(that: VirtualSourceElement): Int = {
+        // x < 0 when this < that; x == 0 when this == that; x > 0 when this > that
+        that match {
+            case VirtualModule(thatName) => name.compareTo(thatName)
+            case _                       => -1
+        }
+    }
+
+    override def classType: ReferenceType = throw new UnsupportedOperationException("module is not a class member")
```
johannesduesing, line 81: This looks a bit shaky .. if i look at the JavaDoc of `VirtualSourceElement.classType` there is no mention that this method may throw an exception. I undestand it works - i looked through the code and saw no relevant usage of this method in the dependency extractor. But i am wondering if either not all `VirtualSourceElement`s have a `classType` or `VirtualModule` should not be a  `VirtualSourceElement`.

errt: That is true. I did not find a better solution, however. I'm open for any ideas. The problem is that `VirtualSourceElement` was designed for the source elements that existed at the time, classes and their members. But now, the dependency extractor also has to handle modules, and modules are not class members (or related to any particular class).

johannesduesing: As discussed, this will be fine for now 👍

johannesduesing (approved):
LGTM 👍

PR #296, opened 2025-07-22
Add requirements to ContextProvider(Key)

errt:
Required (pun intended) to avoid deadlock in computation of the key.


johannesduesing (approved):
This also looks good to me 👍

PR #297, opened 2025-07-22
Remove unused property classes

errt:
Removes `AggregatableValueProperty` which has become obsolete since the introduction of proper call graphs and mostly implementations have been broken since the introduction of contexts.

Also removes the unnecessary `ExplicityNamedProperty` trait.


johannesduesing (dismissed):
This looks good to me, although it now has some conflicts with the base branch. Since that will change the code anyways i guess we can go ahead and not wait for the results of the IT tests - they'll be triggered on develop either way 🤓

johannesduesing (approved):

PR #298, opened 2025-07-24
Catch empty parameters in ReflectionRelatedFieldAccesses.

### LINKED ISSUE #295: Possible empty actualParams in ReflectionRelatedFieldAccessesAnalysis
Ran into an issue where it is possible to get `Some(Seq())` for the actualParams [here](https://github.com/opalj/opal/blob/d48fc34a97bab89afcea70414ec400764a201012/OPAL/tac/src/main/scala/org/opalj/tac/fpcf/analyses/fieldaccess/reflection/ReflectionRelatedFieldAccessesAnalysis.scala#L1058) in ReflectionRelatedFieldAccessesAnalysis when generating a callgraph with AllocationSiteBasedPointsToCallGraphKey. The `Seq.head` call will then give an `class java.lang.ArrayIndexOutOfBoundsException: Index 0 out of bounds for length 0` error.

Based on my understanding, which could be incomplete, this might be fixed with `if (!handleData.isStatic && actualParams.exists(_.nonEmpty)) {`.

errt:
Fixes #295


johannesduesing (approved):
This is looking like a reasonable fix to me 👍

PR #299, opened 2025-07-24
scalafmt update

errt:
Update to the latest release of scalafmt. While not all changes are perfect, they seem to be an overall improvement. Should also possibly improve performance.


johannesduesing (approved):
lgtm 👍

PR #300, opened 2025-08-03
Use IDE ICFG for IFDS

Rkoer:
- Integrate the ICFG from IDE into the IFDS implementation

This pull request depends on PR#240


DEVELOPING_OPAL/validate/src/test/scala/org/opalj/fpcf/ifds/VariableTypeProblem.scala
```diff
@@ -135,7 +135,7 @@ class VariableTypeProblem(project: SomeProject, override val subsumeFacts: Boole
         stmt.astID match {
             case Assignment.ASTID =>
                 // Add facts for the assigned variable.
-                inSet ++ newFacts(statement.method, statement.stmt.asAssignment.expr, statement.index, in)
+                inSet ++ newFacts(statement.method, statement.stmt.asAssignment.expr, statement.pc, in)
```
errt: This doesn't seem correct. In other places, this value is used to check against `definedBy` (e.g., Line 178) which is based on tac indices, not bytecode program counters.

OPAL/ifds/src/main/scala/org/opalj/ifds/IFDSAnalysis.scala
```diff
@@ -226,17 +227,17 @@ case class Statistics(
  *
  * @author Marc Clement
  */
-class IFDSAnalysis[Fact <: AbstractIFDSFact, C <: AnyRef, S <: Statement[_ <: C, _]](
+class IFDSAnalysis[Fact <: AbstractIFDSFact, C <: AnyRef, S, _ICFG <: ICFG[S, C]](
```
errt: Please do not use underscores in generic type parameters

errt (changes requested):

DEVELOPING_OPAL/validate/src/test/scala/org/opalj/fpcf/ifds/VariableTypeProblem.scala
Rkoer: We had a similar discussion in the PR#240. I guess I am gonna rename the `pc` variable of the IDE ICFG implementation to `index`, analogous to the IDE tests. Is that okay with you?

errt: This doesn't seem to be a naming problem. The values are taken from `Stmt.pc` which is a bytecode PC, but compared to `DUVar.definedBy` which is a TAC index.

errt: @Rkoer Can you say something about this?

Rkoer: The involved ICFG and the analyses are all about TAC. The ICFG implementation uses the `CFG.foreachSuccessor` method to get the next PC. The CFG returned from the tacProvider got the PCs converted to indices as far as i understand. So basically all the `JavaStatement.pc` are TAC indices.

errt: I see. So this isn't `Stmt.pc`, but always `JavaStatement.pc`. Then yes, renaming all of these instances to `index` and adapting the ScalaDoc accordingly should be the right way forward. Thank you for the clarification.

errt (approved):
LGTM now

PR #302, opened 2025-08-06
Rename RTJar to JavaBase

errt:
Java 8 is long gone.


johannesduesing (approved):
This is looking good to me 👍 (although i'd disagree with the statement that "java 8 is gone") 🤓

PR #303, opened 2025-08-13
Improve Publications page

errt:
Clarify the papers to cite when using OPAL. Also move the Unimocg paper to the core papers section


johannesduesing (approved):
LGTM 👍

PR #304, opened 2025-08-25
Remove aggregating analyses from config

errt:
Follow up for #297, forgot to remove the deleted analyses (and update the moved one) in the config for the FPCFAnalysesRegistry


johannesduesing (approved):
Seems sensible to me 👍

PR #307, opened 2025-08-28
Util method for reflective instantiation of objects

errt:
We do this in several places, so provide a type-safe util method for it.


johannesduesing (commented):
This seems to be working now. I am, however, asking myself if we need to do anything about this weird dead code error that we observed before. Should we add a hint to the new method's JavaDoc? Or maybe make sure that at least every current usage of `getObjectRefelctively` explicitly sets the type parameter (instead of relying on type inference)?

johannesduesing (approved):
With the latest changes this now looks good to me 👍

PR #311, opened 2025-09-08
Trim class name

errt:
The string must be trimmed to remove the blanks at the start (this used to be the case and was accidentally removed in #307). Fixes https://github.com/opalj/opal/actions/runs/17549630645


johannesduesing (approved):
Looks good 👍

PR #313, opened 2025-09-15
Fix exception for empty parameters

errt:
An exception would occur here if the actualParams are Some(Seq.empty).


maximilianruesch (approved):
That's crazy, this is also part of #214, but we can pull it ahead.

PR #315, opened 2025-09-16
Fix incrementing PC after GOTO insertion

maximilianruesch:
We forgot to increment the PC. The test suite run in #312 was apparently run in an outdated setting.


w-lfchen (approved):
my testing for #312 didn't fail since i most likely forgot to update my sbt config correctly. this should work now.

errt (approved):
LGTM

PR #318, opened 2025-09-18
Add "none" option for CG argument

errt:
To allow for CLI runners that don't need a call graph in all configurations


johannesduesing (approved):
Looks good to me 👍

PR #319, opened 2025-09-18
Fix two CLI runners

errt:
One got broken in #241


johannesduesing (approved):
Yes that looks reasonable, i overlooked that when changing the definition sites - good catch 🤓

PR #320, opened 2025-09-18
Log used ContextProvider

errt:
ContextProviders are a central building block now, track which one is used for better debugging


johannesduesing (commented):
I am not a big fan of printing those memory addresses to the console .. currently the output looks like this:
```
[info][analysis configuration] the ContextProvider used is org.opalj.tac.fpcf.analyses.cg.PropagationBasedTypeIterator@1e141e42
```
I think it would be better to just print the class name. The same thing is currently true for our logging of the scheduling strategy that is selected. What's your opinion on that?

johannesduesing:
Also, to once again bring up the scheduling strategies, maybe we should agree on a consistent wording here? For the scheduling strategy, the logs says:
```
scheduling strategy org.opalj.fpcf.scheduling.MaximumPhaseScheduling$@64469d8 is selected
```
Maybe we want to use the same phrase for the context provider?

errt:
Sure, we can make messages more consistent.

johannesduesing (approved):
This should now cover all relevant cases of context providers with instance information

PR #321, opened 2025-09-18
CLI-Runner smoke test

errt:
A new smoke test that can catch various issues by running all suitable CLI runners


johannesduesing (approved):
Tests are pass now that the other two PRs are merged. Looks good to me, although it does increase the unit test runtime a bit.

PR #323, opened 2025-09-18
Add cleaning to the PropertyStore

fwnter:
Adds a default to cleanup the PropertyStore between phases. Properties to be deleted/kept can be set via optional cli-parameters and the cleanup itself can be disabled.


DEVELOPING_OPAL/tools/src/main/scala/org/opalj/support/info/Immutability.scala
```diff
@@ -54,14 +50,7 @@ import org.opalj.br.fpcf.properties.immutability.TypeImmutability
...
 import org.opalj.bytecode.JRELibraryFolder
-import org.opalj.cli.AnalysisLevelArg
-import org.opalj.cli.ChoiceArg
-import org.opalj.cli.ConfigurationNameArg
-import org.opalj.cli.LibraryArg
-import org.opalj.cli.OutputDirArg
-import org.opalj.cli.ParsedArg
-import org.opalj.cli.PlainArg
-import org.opalj.cli.ThreadsNumArg
+import org.opalj.cli.{AnalysisLevelArg, ChoiceArg, ClearPropertyKeysArg, ConfigurationNameArg, DisableCleanupArg, KeepPropertyKeysArg, LibraryArg, OutputDirArg, ParsedArg, PlainArg, ThreadsNumArg}
```
errt: Do not use multi-imports (disable this in your IntelliJ settings)

OPAL/bi/src/main/scala/org/opalj/bytecode/package.scala
```diff
@@ -117,7 +117,7 @@ package object bytecode {
     lazy val JRELibraryFolder: File = {
         val javaVersion = System.getProperty("java.version")
         if (javaVersion.startsWith("1.")) {
-            val sunBootClassPath = System.getProperties().getProperty("sun.boot.class.path")
+            val sunBootClassPath = System.getProperties.getProperty("sun.boot.class.path")
```
errt, line 120: These changes seem unrelated to the PR

OPAL/common/src/main/scala/org/opalj/cli/ClearPropertyKeysArg.scala
```diff
@@ -0,0 +1,17 @@
+package org.opalj.cli
```
errt: Split package declarations: org.opalj goes into the first line, then one line for each subpackage

OPAL/common/src/main/scala/org/opalj/cli/ClearPropertyKeysArg.scala
```diff
@@ -0,0 +1,17 @@
...
+    override val description: String = "List of Properties to keep at the end of the analysis"
+    override val defaultValue: Option[String] = None
+    override def apply(config: Config, value: Option[String]): Config = {
+        config.withValue(
+            "org.opalj.fpcf.AnalysisScenario.ClearPropertyKeys",
+            ConfigValueFactory.fromAnyRef(value.getOrElse(""))
+        )
+    }
+
+    override def parse(arg: String): String = arg.trim
```
errt: I don't see why this parse method would be necessary

OPAL/common/src/main/scala/org/opalj/cli/ClearPropertyKeysArg.scala
```diff
@@ -0,0 +1,17 @@
+package org.opalj.cli
+
+import com.typesafe.config.{Config, ConfigValueFactory}
+
+object ClearPropertyKeysArg extends ParsedArg[String, String] {
```
errt: Since the description says it is a list of properties, why is the type String, not List[String]?

OPAL/si/src/main/scala/org/opalj/fpcf/PhaseConfiguration.scala
```diff
@@ -4,5 +4,6 @@ package fpcf
 
 case class PhaseConfiguration[A](
     propertyKinds: PropertyKindsConfiguration,
-    scheduled:     List[ComputationSpecification[A]]
+    scheduled:     List[ComputationSpecification[A]],
+    var toDelete:  Set[Int] = Set.empty
```
errt: Avoid `var`s please, especially in case classes like this

OPAL/si/src/main/scala/org/opalj/fpcf/PropertyStore.scala
```diff
@@ -1012,6 +1014,10 @@ abstract class PropertyStore {
             case t: Throwable         => collectAndThrowException(t)
         }
     }
+
+    def setCurrentPhaseConfiguration[A](pc: PhaseConfiguration[A]): Unit = {
```
errt: Can't we include this into `setupPhase`?

OPAL/si/src/main/scala/org/opalj/fpcf/Schedule.scala
```diff
@@ -46,13 +46,17 @@ case class Schedule[A](
...
 
             if (trace) {
                 info("analysis progress", s"setting up analysis phase $id: $configuration")
             }
             time {
                 ps.setupPhase(configuration)
+                ps.setCurrentPhaseConfiguration(
+                    PhaseConfiguration(configuration, css, toDelete)
+                )
+                afterPhaseSetup(configuration)
```
errt, line 67: What's with the duplicate `afterPhaseSetup` here?

OPAL/si/src/main/scala/org/opalj/fpcf/par/PKECPropertyStore.scala
```diff
@@ -472,7 +470,7 @@ class PKECPropertyStore(
         threads.foreach { _.start }
         threads.foreach { _.join }
         if (doTerminate) {
-            if (exception ne null) throw exception;
+            if (exception ne null) throw exception
```
errt: Do not remove these semicola, they are there for a reason (marking early returns)

OPAL/si/src/main/scala/org/opalj/fpcf/par/PKECPropertyStore.scala
```diff
@@ -519,12 +517,19 @@ class PKECPropertyStore(
...
 
             subPhaseId += 1
-
+            clearObsoletePropertyKinds()
             ps(AnalysisKeyId).clear()
         }
 
         idle = true
     }
+    private def clearObsoletePropertyKinds(): Unit = {
```
errt: Add blank line before method declaration

OPAL/si/src/main/scala/org/opalj/fpcf/par/PKECPropertyStore.scala
```diff
@@ -519,12 +517,19 @@ class PKECPropertyStore(
...
+        currentPhaseConf match {
+            case Some(pc) =>
+                pc.toDelete.foreach{ key => ps(key).clear() }
+            case None =>
```
errt: Should this ever be None?

build.sbt
```diff
@@ -607,4 +607,4 @@ ThisBuild / publishMavenStyle.withRank(KeyRanks.Invisible) := true
 ThisBuild / sonatypeCredentialHost := sonatypeCentralHost
 Test / publishArtifact := false
 ThisBuild / publishTo := sonatypePublishToBundle.value
-ThisBuild / pomExtra := MavenPublishing.pomNodeSeq()
+ThisBuild / pomExtra := MavenPublishing.pomNodeSeq()
```
errt: This change seems unrelated to the PR (and also, please keep terminating new lines also in other files)

OPAL/si/src/main/scala/org/opalj/fpcf/scheduling/MaximumPhaseScheduling.scala
```diff
@@ -98,7 +97,39 @@ abstract class MultiplePhaseScheduling extends SchedulingStrategy {
             computePhase(ps, phaseAnalyses, remainingAnalyses)
         }
 
-        schedule
+        if(config.getBoolean(DisableCleanup)) calculateDeletions(schedule, ps, config) else schedule
+    }
+
+    private def calculateDeletions[A](schedule: List[PhaseConfiguration[A]], ps: PropertyStore, config: Config): List[PhaseConfiguration[A]] = {
+        def getSetOrNilForProperty(propertyKey: String) = {
```
errt: Specify return type for the method please

OPAL/si/src/main/scala/org/opalj/fpcf/scheduling/MaximumPhaseScheduling.scala
```diff
@@ -98,7 +97,39 @@ abstract class MultiplePhaseScheduling extends SchedulingStrategy {
...
+            if (!config.hasPath(propertyKey)) Nil
+            else {
+                val rawValue = config.getString(propertyKey).trim
+                if (rawValue.isEmpty) Set.empty[Int]
```
errt: Is this necessary? I suppose `split` would return an empty sequence anyway

OPAL/si/src/main/scala/org/opalj/fpcf/scheduling/MaximumPhaseScheduling.scala
```diff
@@ -98,7 +97,39 @@ abstract class MultiplePhaseScheduling extends SchedulingStrategy {
...
 
-        schedule
+        if(config.getBoolean(DisableCleanup)) calculateDeletions(schedule, ps, config) else schedule
+    }
+
+    private def calculateDeletions[A](schedule: List[PhaseConfiguration[A]], ps: PropertyStore, config: Config): List[PhaseConfiguration[A]] = {
+        def getSetOrNilForProperty(propertyKey: String) = {
+            if (!config.hasPath(propertyKey)) Nil
+            else {
+                val rawValue = config.getString(propertyKey).trim
```
errt: Why all the `trim`s? If the format of the config is properly speficied, there should be nothing to trim.

OPAL/si/src/main/scala/org/opalj/fpcf/scheduling/MaximumPhaseScheduling.scala
```diff
@@ -98,7 +97,39 @@ abstract class MultiplePhaseScheduling extends SchedulingStrategy {
...
+                if (rawValue.isEmpty) Set.empty[Int]
+                else rawValue.split(",").map(_.trim.toInt).toSet
```
errt: We can't specify property keys by their ids, those are not guaranteed to be stable. If you need them specified, you must do so by name. But I still suggest a programmatical way to do it instead of (ab)using the config.

OPAL/si/src/main/scala/org/opalj/fpcf/scheduling/MaximumPhaseScheduling.scala
```diff
@@ -98,7 +97,39 @@ abstract class MultiplePhaseScheduling extends SchedulingStrategy {
...
+            }
+        }
+        val pkToKeep = getSetOrNilForProperty(PropertiesToKeepKey)
+        val pkToDelete = getSetOrNilForProperty(PropertiesToRemoveKey)
+         val properties: mutable.Set[Int] =
```
errt: There's clearly some formatting gone wrong here. Please always run `sbt format` before pushing commits

errt (changes requested):

OPAL/si/src/main/scala/org/opalj/fpcf/PropertyStore.scala
```diff
@@ -531,7 +530,8 @@ abstract class PropertyStore {
         pc: EOptionP[E, P] => InterimEP[E, P]
     ): Unit
 
-    final def setupPhase(configuration: PropertyKindsConfiguration): Unit = {
+    final def setupPhase(configuration: PropertyKindsConfiguration, toDelete: Set[Int]): Unit = {
```
errt, line 563: Consider whether the `toDelete` could be part of the configuration

OPAL/si/src/main/scala/org/opalj/fpcf/PropertyStore.scala
```diff
@@ -531,7 +530,8 @@ abstract class PropertyStore {
...
+        this.currentPhaseToDelete = toDelete
```
errt: Consider whether to move this to the other `setupPhase` method, so it is always set (since this method here is just a forwarder).

OPAL/si/src/main/scala/org/opalj/fpcf/PropertyStore.scala
```diff
@@ -135,8 +135,7 @@ abstract class PropertyStore {
     //
 
     private[this] val externalInformation = new ConcurrentHashMap[AnyRef, AnyRef]()
-
-    var currentPhaseConf: Option[PhaseConfiguration[_]] = None
+    protected[this] var currentPhaseToDelete: Set[Int] = Set.empty
```
errt: Consider to make this consistent with other phase configurations, i.e., an array of Boolean values.

OPAL/si/src/main/scala/org/opalj/fpcf/scheduling/MaximumPhaseScheduling.scala
```diff
@@ -97,38 +102,59 @@ abstract class MultiplePhaseScheduling extends SchedulingStrategy {
...
+        def getSetForProperty(propertyKey: String): Set[Int] = {
+            Option(
+                config.getString(propertyKey)
+            ).toSeq.flatMap(_.split(",")).filter(_.nonEmpty).map(name => keyByName(name).id).toSet
         }
-        val pkToKeep = getSetOrNilForProperty(PropertiesToKeepKey)
-        val pkToDelete = getSetOrNilForProperty(PropertiesToRemoveKey)
-         val properties: mutable.Set[Int] =
-            schedule.iterator.flatMap(_.propertyKinds.propertyKindsComputedInThisPhase.map(key => key.id)).to(mutable.Set)
+        val pkToKeep = getSetForProperty(PropertiesToKeepKey)
```
errt: Still unsure whether this should really be solved via the config. It seems too variable to me. Also, it doesn't allow for a simple programmatic way to set these.

OPAL/common/src/main/scala/org/opalj/cli/ClearPropertyKeysArg.scala
```diff
@@ -0,0 +1,19 @@
...
+    override val description: String = "List of Properties to keep at the end of the analysis"
+    override val defaultValue: Option[List[String]] = None
+    override def apply(config: Config, value: Option[List[String]]): Config = {
+        config.withValue(
+            "org.opalj.fpcf.AnalysisScenario.ClearPropertyKeys",
+            ConfigValueFactory.fromAnyRef(value.getOrElse(""))
+        )
+    }
+
+    override def parse(arg: List[String]): List[String] = ???
```
errt: There seems to be implementation missing here. Probably, this should return something like a `List[PropertyKey]`, having parsed the Strings to PropertyKeys.

OPAL/si/src/main/scala/org/opalj/fpcf/scheduling/MaximumPhaseScheduling.scala
```diff
@@ -98,7 +102,60 @@ abstract class MultiplePhaseScheduling extends SchedulingStrategy {
...
-        schedule
+        if (!config.getBoolean(DisableCleanup)) calculateDeletions(schedule, ps, config) else schedule
+    }
+
+    private def calculateDeletions[A](
+        schedule: List[PhaseConfiguration[A]],
+        ps:       PropertyStore,
+        config:   Config
+    ): List[PhaseConfiguration[A]] = {
+        def keyByName(name: String): SomePropertyKey = {
```
errt: We might want to have a more efficient way of turning names into keys implemented in `PropertyKey`, e.g., a map from name to key.

OPAL/si/src/main/scala/org/opalj/fpcf/scheduling/MaximumPhaseScheduling.scala
```diff
@@ -98,7 +102,60 @@ abstract class MultiplePhaseScheduling extends SchedulingStrategy {
...
+                    (properties -- pkToKeep) -- alreadyDeleted
+                } else
+                    (properties -- usedDependencies) -- alreadyDeleted
+            }
+
+            alreadyDeleted ++= remaining
+            index -= 1
+        }
+        modifiedSchedule(lastIndex) = modifiedSchedule(lastIndex).copy(
+            toDelete = (modifiedSchedule(lastIndex).toDelete ++ pkToDelete) -- pkToKeep
```
errt: This looks like it would only compute PKs to be deleted after the last phase, which is clearly not what we want

OPAL/si/src/main/scala/org/opalj/fpcf/scheduling/MaximumPhaseScheduling.scala
```diff
@@ -98,7 +102,60 @@ abstract class MultiplePhaseScheduling extends SchedulingStrategy {
             computePhase(ps, phaseAnalyses, remainingAnalyses)
         }
 
-        schedule
+        if (!config.getBoolean(DisableCleanup)) calculateDeletions(schedule, ps, config) else schedule
```
errt: As we allow users to specify PKs to be deleted (even after the last phase), this probably has to be applied to SinglePhaseScheduling as well, i.e., can't be in this class.

OPAL/common/src/main/scala/org/opalj/cli/ClearPropertyKeysArg.scala
fwnter: It can't be parsed to PropertyKeys since those aren't accessible within cli

errt: Oh, these args should not be in the `common` subproject anyway. They belong to the static analysis infrastructure in `si`, just as the rest of your implementation. Cf. `org.opalj.fpcf.PropertyStoreBasedCommandLineConfig`

fwnter: Ahh, got it, thanks!

OPAL/si/src/main/scala/org/opalj/fpcf/PropertyStore.scala
```diff
@@ -530,13 +543,13 @@ abstract class PropertyStore {
...
-    final def setupPhase(configuration: PropertyKindsConfiguration, toDelete: Set[Int]): Unit = {
-        this.currentPhaseToDelete = toDelete
+    final def setupPhase(configuration: PropertyKindsConfiguration): Unit = {
         setupPhase(
             configuration.propertyKindsComputedInThisPhase,
             configuration.propertyKindsComputedInLaterPhase,
             configuration.suppressInterimUpdates,
-            configuration.collaborativelyComputedPropertyKindsFinalizationOrder
+            configuration.collaborativelyComputedPropertyKindsFinalizationOrder,
+            toDelete = Set.empty
```
errt: I don't get why this is fixed to empty. It could also be provided as a (may default empty) argument to the method.

OPAL/si/src/main/scala/org/opalj/fpcf/PropertyStore.scala
```diff
@@ -500,6 +500,19 @@ abstract class PropertyStore {
...
+        var index = 0
+        while (index < mask.length) {
+            if (mask(index)) {
+                clearSlot(index)
+            }
+            index += 1
+        }
+    }
+
+    protected def clearSlot(id: Int): Unit
```
errt: This name seems a little unintuitive. Also, please document your methods.

OPAL/si/src/main/scala/org/opalj/fpcf/scheduling/Cleanup.scala
```diff
@@ -0,0 +1,71 @@
...
+     * @param config Config to be read from
+     * @return A new [[CleanupSpec]]
+     */
+    def fromConfig(config: Config): CleanupSpec = {
+        def getSetForProperty(propertyKey: String): Set[Int] = {
+            Option(
+                config.getString(propertyKey)
+            ).toSeq.flatMap(_.split(",")).filter(_.nonEmpty).map(PropertyKey.idByName).toSet
+        }
+        val toKeep = getSetForProperty(PropertiesToKeepKey)
```
errt: Why are these still handled via the config?

OPAL/si/src/main/scala/org/opalj/fpcf/scheduling/Cleanup.scala
```diff
@@ -0,0 +1,71 @@
...
+        ps:       PropertyStore,
+        spec:     CleanupSpec
+    ): List[PhaseConfiguration[A]] = {
+        if (spec.disable) return schedule
+
+        val producedInAllPhases: Set[Int] =
+            schedule.iterator.flatMap(_.propertyKinds.propertyKindsComputedInThisPhase.map(_.id)).toSet
+        val neededLater = Array.fill[Set[Int]](schedule.size + 1)(Set.empty)
+        var index = schedule.size - 1
+        var acc: Set[Int] = Set.empty
```
errt: Use self-explaining variable names

OPAL/si/src/main/scala/org/opalj/fpcf/scheduling/Cleanup.scala
```diff
@@ -0,0 +1,71 @@
...
+     * Calculates the properties to be safely removed inbetween phases. Returns an unmodified schedule if cleanup is disabled
+     */
+    def withPerPhaseCleanup[A](
+        schedule: List[PhaseConfiguration[A]],
+        ps:       PropertyStore,
+        spec:     CleanupSpec
+    ): List[PhaseConfiguration[A]] = {
+        if (spec.disable) return schedule
+
+        val producedInAllPhases: Set[Int] =
```
errt: Rather in any phase?

OPAL/si/src/main/scala/org/opalj/fpcf/PropertyKey.scala
```diff
@@ -111,6 +111,10 @@ object PropertyKey {
         create(name, fpc)
     }
 
+    lazy val byName: Map[String, SomePropertyKey] = {
```
errt: Are you sure this is guaranteed to work? I.e., that this lazy val is only computed after all property keys have been created?

fwnter: Since I wasn't sure about the guaranteed-part (and am not willing to put my hand into the fire), I changed it to `propertyKeys.find(..)` within the `getByName()`

DEVELOPING_OPAL/tools/src/main/scala/org/opalj/support/info/Immutability.scala
```diff
@@ -132,7 +135,10 @@ object Immutability extends ProjectsAnalysisApplication {
             analysisLevelArg !,
             ignoreLazyInitializationArg !,
             ConfigurationNameArg !,
-            EscapeArg
+            EscapeArg,
+            DisableCleanupArg,
```
errt: These should probably be added to the `PropertyStoreBasedCommandLineConfig`

OPAL/si/src/main/scala/org/opalj/fpcf/PropertyStore.scala
```diff
@@ -529,12 +550,13 @@ abstract class PropertyStore {
         pc: EOptionP[E, P] => InterimEP[E, P]
     ): Unit
 
-    final def setupPhase(configuration: PropertyKindsConfiguration): Unit = {
+    final def setupPhase(configuration: PropertyKindsConfiguration, toDelete: Set[Int]): Unit = {
```
errt, line 563: Should the `toDelete` here get a default of `Set.empty`? Or is that not possible because of the other `setupPhase` method? (In which case a second forwarder taking only a `PropertyKindsConfiguration` might be an option)

OPAL/si/src/main/scala/org/opalj/fpcf/PropertyStore.scala
```diff
@@ -584,6 +607,11 @@ abstract class PropertyStore {
             "illegal self dependency"
         )
 
+        val mask = currentPhaseDeletionMask
```
errt: Document this as a step of its own and move it to a suitable place in this method (probably near the end).

OPAL/si/src/main/scala/org/opalj/fpcf/Schedule.scala
```diff
@@ -43,16 +45,26 @@ case class Schedule[A](
...
+            val (phase, id) = batchId
+            val configuration = phase.propertyKinds
+            val css = phase.scheduled
 
             if (trace) {
                 info("analysis progress", s"setting up analysis phase $id: $configuration")
             }
             time {
-                ps.setupPhase(configuration)
+                ps.setupPhase(
```
errt: You can probably use the two-argument overload of `setupPhase` here

OPAL/si/src/main/scala/org/opalj/fpcf/PropertyStore.scala
fwnter: Added a second forwarder to resolve this

OPAL/si/src/main/scala/org/opalj/fpcf/Schedule.scala
```diff
@@ -56,6 +56,9 @@ case class Schedule[A](
 
             if (trace) {
                 info("analysis progress", s"setting up analysis phase $id: $configuration")
+                if (phase.toDelete.nonEmpty) {
+                    info("cleanup", s"to be deleted after this phase: " + phase.toDelete.map(PropertyKey.name))
```
errt: Put this into "analysis progress"

OPAL/si/src/main/scala/org/opalj/fpcf/PropertyStore.scala
```diff
@@ -674,6 +669,13 @@ abstract class PropertyStore {
             suppressInterimUpdates,
             finalizationOrder
         )
+
+        // Step 6
```
errt: Should probably be before current Step 5 and also pass the deletions to `newPhaseInitialized`

fwnter: As in extending the `newPhaseInitialized` with the deletionMask?

fwnter: Check

OPAL/si/src/main/scala/org/opalj/fpcf/scheduling/Cleanup.scala
```diff
@@ -7,12 +7,13 @@ package scheduling
...
+ * @param enable Allows the cleanup to be enabled since it is off by default
  */
 final case class CleanupSpec(
-    keep:    Set[Int] = Set.empty,
-    clear:   Set[Int] = Set.empty,
-    disable: Boolean  = false
+    keep:                  Set[Int] = Set.empty,
+    clear:                 Set[Int] = Set.empty,
+    enable:                Boolean  = false,
+    cleanupAfterLastPhase: Boolean  = false
```
errt: This parameter is not documented. I also don't see it used anywhere?

OPAL/si/src/main/scala/org/opalj/fpcf/scheduling/Cleanup.scala
```diff
@@ -24,13 +25,13 @@ object Cleanup {
      * Creates a [[CleanupSpec]] from given [[PropertyKey]]-names to keep and/or to clear. Also allows disabling the cleanup by setting 'disable' to 'true'.
      * @param keep Names of PropertyKeys to be kept after the analyses
      * @param clear Names of PropertyKeys to be removed after the analyses
-     * @param disable Setting this to 'true' disables the cleanup inbetween the phases
+     * @param enable Setting this to 'true' enables the cleanup inbetween the phases
      * @return A new [[CleanupSpec]]
      */
-    def fromArgs(keep: Set[String], clear: Set[String], disable: Boolean): CleanupSpec = {
+    def fromArgs(keep: Set[String], clear: Set[String], enable: Boolean): CleanupSpec = {
```
errt: Why is this called `fromArgs` and an instance method? Should it not rather be an `apply` method on a companion object?

OPAL/si/src/main/scala/org/opalj/fpcf/scheduling/Cleanup.scala
```diff
@@ -56,9 +57,17 @@ object Cleanup {
             index -= 1
         }
 
+        var alreadyCleared = Set.empty[Int]
+        val lastIndex = schedule.size - 1
         schedule.indices.iterator.map { index =>
             val producedHere = schedule(index)
-            val toDelete = ((producedInAnyPhase -- neededLater(index)) -- spec.keep) union spec.clear
+            val toDelete = if (index == lastIndex) {
```
errt, line 85: Maybe add some comments to document the algorithm.

OPAL/si/src/main/scala/org/opalj/fpcf/scheduling/Cleanup.scala
fwnter: The idea behind it was that since I'm transforming the keep/clear at this point, I'm creating it `fromArgs`. But you're right, moving it to a dedicated companion object makes more sense here.

PR #326, opened 2025-09-22
Adapt purity tests to Java 25 changes

errt:
With the Java 25 javac, nested class constructors call Objects.requireNonNull which creates a NullPointerException, which uses its own implementation of fillInStackTrace


johannesduesing (dismissed):
I fixed a minor issue in the assertions and IT tests are now pass - looks good to me 👍

OPAL/tac/src/main/scala/org/opalj/tac/fpcf/analyses/TACAIBasedAnalysisState.scala
```diff
@@ -26,7 +26,7 @@ trait TACAIBasedAnalysisState[TheContextType <: Context]
     def callContext: ContextType
 
     protected[this] var _tacDependee: EOptionP[Method, TACAI]
-    assert(_tacDependee.hasUBP)
+    assert(_tacDependee == null || _tacDependee.hasUBP)
```
errt: This looks dangerous. When _tacDependee is null, methods like the ones below will throw exceptions.

johannesduesing: Yeah you're right, we should probably keep that assertion. However, the reason that we see the error in the integration tests is actually [this line](https://github.com/opalj/opal/blob/b09f6fa572974e164eabf87203160342ac759e81/OPAL/tac/src/main/scala/org/opalj/tac/fpcf/analyses/pointsto/NewInstanceAnalysis.scala#L128-L129) in the `NewInstanceAnalysis` which i do not know much about - this analysis explicitly introduces `null` as the `_tacDependee`. I guess it would be best to change that to `FinalEP(null, NoTACAI)` like we did for the configured native methods analysis - would you agree?

johannesduesing: Addition to that: It seems like many analyses explicitly pass `null` as the `_tacDependee` to the `PointsToAnalysisState`, i am very confused how this problem did not show up earlier. At least the `NewInstanceAnalysis`, `ReflectionAllocationAnalysis`, `SerializationAllocationAnalysis` and the `TamiFlexPointsToClassGetMemberAnalysis` do so.

errt: Probably because nobody tested with assertions...

johannesduesing (approved):
Assertion is back in place, and no more null values are passed to the `PointsToAnalysisState` - lgtm 👍

PR #331, opened 2025-10-02
Move constant ClassType.apply usages to constant values

### LINKED ISSUE #228: Flushing Type Cache leads to Exceptions in AI
I've encountered an issue with the "Flush Type Cache" functionality introduced in #73 - this is closely related to what we found when we initially added that feature (see #75). I'm aware that we likely cannot do much about this, but i wanted to document the issue either way.

**Problem**
I implemented a large-scale analysis that analyzes many projects in a row and flushes caches in-between to avoid filling up the heap space. The analysis runs `project.get(ComputeTACAIKey)` for every project. Occasionally this fails with the following exception

```
org.opalj.ai.InterpretationFailedException$$anon$1: InterpretationFailedException(
        domain=org.opalj.ai.domain.l1.DefaultDomainWithCFGAndDefUse@5720c48d with TheMethod(org.apache.commons.logging.LogFactory{ private static boolean implementsLogFactory(java.lang.Class) }),
        ai=class org.opalj.ai.BaseAI$,
        cause=org.opalj.ai.DomainException: unsupported Class { java.lang.Class forName(java.lang.String,boolean,java.lang.ClassLoader) },
        pc=33,
        operands=List(_ <: java.lang.ClassLoader[↦7;refId=104], int = 0, String("org.apache.commons.logging.LogFactory")[@29;refId=124]),
        registers=(0,_ <: java.lang.Class[↦-1;refId=102]),(1,int = 0),(2,_ <: java.lang.ClassLoader[↦7;refId=104]),(3,null)
)
```

The reason is that the method under analysis calls `Class.forName(String, boolean, ClassLoader)`, and the following match statement in `org.opalj.ai.domain.l1.ClassValues` fails although it does have a specific case for this particular method descriptor.
```
methodDescriptor match {
    case `forName_String`                           => simpleClassForNameCall(pc, value)
    case `forName_String_boolean_ClassLoader`       => simpleClassForNameCall(pc, value)
    case `forName_String_Class`                     => simpleClassForNameCall(pc, value)
    case `forName_String_boolean_ClassLoader_Class` => simpleClassForNameCall(pc, value)
    case _ =>
        throw new DomainException(
            s"unsupported Class { ${methodDescriptor.toJava("forName")} }"
        )
}
```
This is turn is caused by the same issue we had with the test executions in #73 - `ObjectType("java/lang/ClassLoader")` is created once (see code below) and always used for comparisons in the above match-statement. However, after flushing the type cache, the next project under analysis will create a new `ObjectType` instance for the same FQN, yielding a different ID. Since comparison is ID-based, OPAL will conclude that the two instances represent different types and not match the appropriate case.
```
private object ClassValues  {
[..]
  final val forName_String_boolean_ClassLoader = {
      MethodDescriptor(
          ArraySeq(ObjectType.String, BooleanType, ObjectType("java/lang/ClassLoader")),
          ObjectType.Class
      )
  }
[..]
}
```

**Fixing this Issue**
I'm not sure we want or can do something about this. The ultimate reason is that `ObjectType("java/lang/Classloader")` is not one of the predefined OTs (like `ObjectType.String` for example) and will not be kept when flushing the cache. The only way of dealing with this would be to collect **all OTs** that OPAL uses internally and that are not predefined (i.e., created via `ObjectType("<FQN>")`), and make them a predefined constant. A quick search yields over 100 occurrances of that constructor being used in OPAL, making this a major change that is likely not worth it. For me, re-running the failed projects one at a time solves the issue (since there is no cache-flushing involved), however this does not really scale and defeats the purpose of flushing the caches.

johannesduesing:
This PR closes #228. To avoid issues with flushing the type cache, this PR moves most `ClassType` constants that are used internally by OPAL analyses into the set of predefined types, so they are never flushed and IDs stay constant for comparisons. The following constants have *NOT* been replaced (i am open to discussion on whether this makes sense):
* All usages in test files
* Definitions in `tac.fpcf.analyses.purity.DomainSpecificRater` - this defines some very specific logging related types that (i think) are not general enough to be included in the set of predefined types
* All constant definitions in Hermes - many of which are also rather specific, plus Hermes is going to be separate from the core OPAL project soon anyway.
* Dummy type definitions seen mainly in tools / demos / dependency extraction - we do not want invalid `ClassType` objects to be predefined


OPAL/ai/src/main/scala/org/opalj/ai/fpcf/analyses/LBFieldValuesAnalysis.scala
```diff
@@ -437,9 +437,9 @@ object FieldValuesAnalysis {
     // IMPROVE Move to configuration file.
     final val ignoredFields: Map[ClassType, Set[String]] = Map(
         ClassType.System -> Set("in", "out", "err"),
-        ClassType("java/net/InterfaceAddress") -> Set("address"),
-        ClassType("java/util/concurrent/FutureTask") -> Set("runner"),
-        ClassType("java/nio/channels/SelectionKey") -> Set("attachment")
+        ClassType.InterfaceAddress -> Set("address"),
```
errt: These seem quite specific

OPAL/apk/src/main/scala/org/opalj/apk/analyses/ContextRegisteredReceiversAnalysis.scala
```diff
@@ -24,11 +24,8 @@ import org.opalj.value.ValueInformation
  * @author Nicolas Gross
  */
 object ContextRegisteredReceiversAnalysis {
+
```
errt, line 27: I'm not sure we should move Android-specific things into OPAL's core modules

OPAL/br/src/main/scala/org/opalj/br/analyses/cg/EntryPointFinder.scala
```diff
@@ -153,7 +153,7 @@ trait ApplicationWithoutJREEntryPointsFinder extends ApplicationEntryPointsFinde
         }.filterNot { ep =>
             // The WrapperGenerator class file is part of the rt.jar in 1.7., but is in the
             // default package.
-            ep.declaringClassType == ClassType("WrapperGenerator")
+            ep.declaringClassType == ClassType.WrapperGenerator
```
errt: This seems very specific

OPAL/br/src/test/scala/org/opalj/br/ClassTypeTest.scala
```diff
@@ -18,7 +18,7 @@ class ClassTypeTest extends AnyFunSuite {
     }
 
     test("toJavaClass method") {
-        val ot2 = ClassType("java/lang/Object")
+        val ot2 = ClassType.Object
         val ot3 = ClassType("java/lang/String")
```
errt: Why have these instances of String and List not been replaced?

OPAL/tac/src/main/scala/org/opalj/tac/fpcf/analyses/cg/DoPrivilegedCGAnalysis.scala
```diff
@@ -160,137 +160,132 @@ class DoPrivilegedCGAnalysis private[cg] (
     def analyze(p: SomeProject): PropertyComputationResult = {
         var analyses: List[DoPrivilegedMethodAnalysis] = Nil
 
-        val accessControllerType = ClassType("java/security/AccessController")
-        val privilegedActionType = ClassType("java/security/PrivilegedAction")
-        val privilegedExceptionActionType = ClassType("java/security/PrivilegedExceptionAction")
-        val accessControlContextType = ClassType("java/security/AccessControlContext")
-        val permissionType = ClassType("java/security/Permission")
-        val permissionsArray = ArrayType(permissionType)
+        val permissionsArray = ArrayType(ClassType.Permission)
```
errt: Some of these seem very specific. At least consider prefixing the class name with the package?

DEVELOPING_OPAL/demos/src/main/scala/org/opalj/ai/CipherGetInstanceStrings.scala
```diff
@@ -62,14 +56,14 @@ object CipherGetInstanceStrings extends ProjectsAnalysisApplication {
             val result = BaseAI(m, new DefaultDomainWithCFGAndDefUse(project, m))
             val code = result.domain.code
             for {
-                PCAndInstruction(pc, INVOKESTATIC(Cipher, false, "getInstance", _)) <- code
+                PCAndInstruction(pc, INVOKESTATIC(ClassType.Cipher, false, "getInstance", _)) <- code
```
errt: These seem very specific. At least consider prefixing the class name with the package?

errt (approved):
LGTM

PR #333, opened 2025-10-08
Add argument to provide config values

### LINKED ISSUE #328: Provide a Config Argument for CLI Runners
It would be nice to have a generic argument for CLI runners that can override any OPAL config value. Something along the lines of Java's`-D<property>=<value>`. It seems like Scallop should support this out of the box.

errt:
Adds a generic command-line argument that allows overwriting/setting arbitrary config values.

Syntax is `-Ckey1=value1,key2=value2` (the comma can optionally be replaced with a blank and a blank after the initial 'C' is also valid). Arbitrarily nested lists and objects are possible as values, but any `,` or `=` must be escaped by `\` for the argument to be parsed correctly.

I opted for `-C` for the config, please comment if you would prefer a different character. Instead of a single character, we could also have this as a long version, e.g., `--config key1=value1 key2=value2` if that seems more intuitive.

Closes #328


johannesduesing (approved):
I do like the `-C` single character decision, i think we should keep it as it is 👍

PR #335, opened 2025-10-09
Fix index for bootstrap method

errt:
Since bootstrap arguments can be DynamicConstants, adding the bootstrap arguments to the constant pool can also add additional bootstrap methods.


johannesduesing (approved):
LGTM 👍

PR #336, opened 2025-10-09
Remove duplicate time unit output

errt:
The `analysisTime` object's toString already prints the unit.


johannesduesing (approved):
good find 👍

PR #342, opened 2025-10-21
Add inline function for assertions that are elided in production builds

errt:
Since Scala 3 removed the `-Xdisable-assertions` compiler option (and other related features such as `-Xelide-below`), we need a new way to have assertions without runtime impact. This solution is somewhat hacky but seems to work reasonably well.

### DIFF
--- a/OPAL/common/src/main/scala/org/opalj/util/package.scala
+++ b/OPAL/common/src/main/scala/org/opalj/util/package.scala
@@ -12,6 +12,7 @@ import scala.util.Using
 import com.typesafe.config.Config
 import com.typesafe.config.ConfigRenderOptions
 
+import org.opalj.ReleaseFlags.elideAssertions
 import org.opalj.log.GlobalLogContext
 import org.opalj.log.LogContext
 import org.opalj.log.OPALLogger
@@ -136,6 +137,22 @@ package object util {
                 error(category, "Reflected object is invalid", cce)
                 None
         }
+    }
+
+    /**
+     * Assertion that is elided in production builds, controlled by [[ReleaseFlags.elideAssertions]] which is
+     * automatically rewritten for non-SNAPSHOT builds.
+     */
+    inline def elidedAssert(inline assertion: => Boolean): Unit = {
+        inline if (!elideAssertions) assert(assertion)
+    }
+
+    /**
+     * Assertion that is elided in production builds, controlled by [[ReleaseFlags.elideAssertions]] which is
+     * automatically rewritten for non-SNAPSHOT builds.
+     */
+    inline def elidedAssert(inline assertion: => Boolean, inline message: String): Unit = {
+        inline if (!elideAssertions) assert(assertion, message)
     }
 
 }
--- a/build.sbt
+++ b/build.sbt
@@ -199,7 +199,17 @@ lazy val `Common` = (project in file("OPAL/common"))
     .settings(
         name := "Common",
         Compile / doc / scalacOptions := Opts.doc.title("OPAL-Common"),
-        libraryDependencies ++= Dependencies.common(scalaVersion.value)
+        libraryDependencies ++= Dependencies.common(scalaVersion.value),
+        // This tasks flips all `false` flags in the ReleaseFlags file to `true` when compiling production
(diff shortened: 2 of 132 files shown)


OPAL/common/src/main/scala/org/opalj/Assertions.scala
```diff
@@ -0,0 +1,12 @@
+/* BSD 2-Clause License - see OPAL/LICENSE for details. */
+package org.opalj
+
+/**
+ * Defines an `inline val` that controls whether assertions ([[org.opalj.util.elidedAssert]]) are to be elided
+ * during compilation. This file is automaticall rewritten when compiling non-SNAPSHOT builds to enforce elision of
+ * assertions.
```
maximilianruesch:
```suggestion
 * Defines an `inline val` that controls whether assertions ([[org.opalj.util.elidedAssert]]) are to be elided
 * during compilation. This file is automatically rewritten when compiling non-SNAPSHOT builds to enforce elision of
 * assertions.
```

OPAL/common/src/main/scala/org/opalj/util/package.scala
```diff
@@ -138,4 +139,11 @@ package object util {
         }
     }
 
+    /**
+     * Assertion that is elided in production builds.
+     */
```
maximilianruesch, line 145: Perhaps add a reference to where compile time rewriting happens (or even better, point to the `Assertions` object and reference it there)

build.sbt
```diff
@@ -199,7 +199,15 @@ lazy val `Common` = (project in file("OPAL/common"))
     .settings(
         name := "Common",
         Compile / doc / scalacOptions := Opts.doc.title("OPAL-Common"),
-        libraryDependencies ++= Dependencies.common(scalaVersion.value)
+        libraryDependencies ++= Dependencies.common(scalaVersion.value),
+        Compile / sourceGenerators += Def.task {
+            if (!(ThisBuild / version).value.endsWith("-SNAPSHOT")) {
+                val file = (Compile / sourceDirectory).value / "scala" / "org" / "opalj" / "Assertions.scala"
+                IO.write(file, assertionsFileContent.replace("= false", "= true"))
```
maximilianruesch: Perhaps restrict this to the variable `elideAssertions`, such that adding further constants is unaffected by this.

build.sbt
```diff
@@ -551,6 +559,18 @@ compile := {
...
+val assertionsFileContent = IO.read(assertionsFile)
+lazy val reenableAssertions = taskKey[Unit]("Re-enables assertions")
+
+reenableAssertions := {
+    if (!(ThisBuild / version).value.endsWith("-SNAPSHOT")) {
+        IO.write(assertionsFile, assertionsFileContent)
+        streams.value.log.info("Enabling assertions")
+    }
+}
+reenableAssertions := (reenableAssertions triggeredBy (Common / Compile / compile)).value
```
maximilianruesch: Perhaps add an explaining comment and cross reference this with the above addition

maximilianruesch (dismissed):
It indeed seems that they removed the feature, even though it is still referenced in the docs (even for Scala 3.7.3). The annotation will also be deprecated soon (with 3.8.0): https://github.com/scala/scala3/issues/22754

The documentation recommends this approach (minus the compile-time rewriting but ahh well)

build.sbt
errt: I didn't exactly to allow adding further, similarly rewritten constants.

maximilianruesch: Then perhaps this fact should be referenced in the `Assertions` file, so that people are aware that *every* variable is flipped at compile time when they add new ones.

johannesduesing:
I'd agree with the review comments from @maximilianruesch , otherwise i can only report that this approach is working fine on my (windows) machine as well 👍

johannesduesing (changes requested):
All that's changed is looking good to me, we should, however, also revert the changes to our GitHub CI that were done [to enable assertions](https://github.com/opalj/opal/commit/cfc7e27589fc9b30e53eba3dbe6b66383ef5e8f4) - those are no longer valid (or needed).

johannesduesing (approved):
LGTM now 👍

PR #343, opened 2025-10-21
Remove BugPicker

errt:
BugPicker has not been part of OPAL's builds since 2017 and has no real use case anymore.


johannesduesing (approved):
Looks good to me, i did not find any reamining references to BugPicker 👍

PR #344, opened 2025-10-21
Remove Hermes from OPAL

errt:
Hermes will be moved to a repository of its own to simplify OPAL maintenance.


johannesduesing (changes requested):
I found a few remaining occurrances of Hermes, some of which might be on purpose - i still wanted to briefly discuss this:
* There are two files `mapping.hermes.txt` and `results.hermes.txt` in the `.gitignore`
* Hermes Documentation is still part of the website - i guess this is intentional and makes sense to me
* `bi/src/test/fixtures-java/projects/metrics.fanin_fanout/readme.md` explicitly mentiones the path `DEVELOPING_OPAL/tools/.../Hermes`
* The `InvokedynamicRewritingExecutionTest` explicitly executes Hermes via an in-memory class loader - this should not be part of OPALs IT tests anymore, should it?
* `DEVELOPING_OPAL/validate/src/it/resources` contains a fixture file called `hermes-test-fixtures.json`

errt:
Thanks!

- `There are two files mapping.hermes.txt and results.hermes.txt in the .gitignore`
  - I removed them since they don't exist for me so might be outdated anyway
- `Hermes Documentation is still part of the website - i guess this is intentional and makes sense to me`
  - Yes, I kept that intentionally. Hermes is still part of the OPAL family, just maintained separately. We can think of changing the details when we update the website.
- `bi/src/test/fixtures-java/projects/metrics.fanin_fanout/readme.md explicitly mentiones the path DEVELOPING_OPAL/tools/.../Hermes`
  - I am unsure about this file. I think it refers to https://github.com/opalj/opal/blob/91d8f65a10b45a05a63b42f40feb42ead36a6ab8/TOOLS/hermes/src/main/scala/org/opalj/hermes/queries/FanInFanOut.scala but it used to mention a FanInFanOutTest that I could not find in the history, and no other tests for Hermes as well. I think it may be/have been a test fixture for testing that FanInFanOut class, though.
- `The InvokedynamicRewritingExecutionTest explicitly executes Hermes via an in-memory class loader - this should not be part of OPALs IT tests anymore, should it?`
  - Since this seems to execute it from `classfiles/OPAL-MultiJar-SNAPSHOT-01-04-2018.jar`, I think we can keep it.
- `DEVELOPING_OPAL/validate/src/it/resources contains a fixture file called hermes-test-fixtures.json`
  - This is referenced by the InvokedynamicRewritingExecutionTest, so see above.

johannesduesing (approved):
With the latest discussion and changes, this looks good to me 👍

PR #345, opened 2025-10-21
Assorted improvements

errt:
Minor code improvements, mostly from IDE suggestions


johannesduesing (approved):
With the latest test fix, this looks good to me 👍 
I did have a look into the errors during documentation generation, specifically `could not resolve a member for the given link query` - it seems that this can be fixed by using fully qualified names when referencing classes in ScalaDoc comments, i.e., chaning the following
```
/**
 * A scheduler for the reachable method [[SystemPropertiesAnalysis]] that is triggered on the [[Callers]] property.
```
to
```
/**
 * A scheduler for the reachable method [[org.opalj.tac.fpcf.analyses.systemproperties.SystemPropertiesAnalysis]] that is triggered on the [[org.opalj.br.fpcf.properties.cg.Callers]] property.
```
Fixed two of those errors. In total it should account for many of the 314 documentation warnings, but certainly not all of them. I don't know if you want to make that part of this PR or not, just thought i would mention it.

errt:
I thought we tried this before with no success, but it seems to be the correct answer. I wouldn't include it in this PR, but we can certainly try it in another one. Although I'm not too happy that this will create very long links kind of unnecessarily.

PR #346, opened 2025-11-11
Update it:test reference files for fixed typos

errt:
In #345, there were two typos fixed in source code files used for this test that require replacement of the reference files.


johannesduesing (approved):
lgtm

PR #347, opened 2025-11-12
Improve the interface and documentation of the Project class

errt:
The usability of the `Project` class is of great importance given its centralility in OPAL.


OPAL/br/src/main/scala/org/opalj/br/analyses/Project.scala
```diff
@@ -86,74 +86,95 @@ import org.opalj.util.PerformanceEvaluation.time
  *         tools such as IDEs. E.g., in Eclipse `IResource`'s are used to identify the
  *         location of a resource (e.g., a source or class file.)
  *
- * @param  logContext The logging context associated with this project. Using the logging
- *         context after the project is no longer referenced (garbage collected) is not
- *         possible.
+ * @param all[Project|Library]ClassFiles The class files associated with this project, defined in the analyzed project
```
johannesduesing: For me, this syntax produces ScalaDoc warnings - at least within IntelliJ. Maybe write the actual parameter names and accept some duplication within the textual explanations?

OPAL/br/src/main/scala/org/opalj/br/analyses/Project.scala
```diff
@@ -1484,20 +1542,55 @@ object Project {
...
 
-    def apply(file: File, logContext: LogContext, config: Config): Project[URL] = {
+    /**
+     * Creates a `Project` from a class file, jar file, jmod file, or directory containing class-, jar-, and jmod files.
+     *
+     * The global log context will be used for logging messages.
+     *
+     * The given configuration will be used for setting up the project.
+     */
+    def apply(file: File, config: Config): Project[URL] = {
```
johannesduesing, line 1604: Here and afterwards - do we not want to use proper `@param` tags for the (highly relevant) `Project.apply` methods?

johannesduesing (changes requested):
There are two minor issues that i'd like to discuss before we merge this PR.

OPAL/br/src/main/scala/org/opalj/br/analyses/Project.scala
errt: Agreed, even if it produces many duplications.

OPAL/br/src/main/scala/org/opalj/br/analyses/Project.scala
errt: Good point. Done!

johannesduesing (approved):
Looking good, thanks! 🤓

PR #352, opened 2025-12-17
Remove settings.sbt duplicating scalac.options

errt:
These settings are loaded from scala.options anyway where they are configurable.


johannesduesing (approved):
Makes sense, lgtm

PR #355, opened 2025-12-17
Adopt Configuration Explorer Template to new ScalaDoc Version

johannesduesing:
This PR adapts the HTML template used by the *ConfigurationExplorer* to match the apperance of HTML files generated by ScalaDoc 3. This became necessary, as the libraries and CSS scripts used by ScalaDoc changed significantely between both major releases. Also, this PR modifies the build configuration so the configuration explorer is execuded when `sbt unidoc` is run.


TOOLS/ce/src/main/resources/template.html
```diff
@@ -1,58 +1,123 @@
...
+    <script type="text/javascript" src="scripts/common/component.js" defer="true"></script>
+    <script type="text/javascript" src="scripts/common/utils.js" defer="true"></script>
+    <script type="text/javascript" src="scripts/components/FilterBar.js" defer="true"></script>
+    <script type="text/javascript" src="scripts/components/DocumentableList.js" defer="true"></script>
+    <script type="text/javascript" src="scripts/components/Input.js" defer="true"></script>
+    <script type="text/javascript" src="scripts/components/FilterGroup.js" defer="true"></script>
+    <script type="text/javascript" src="scripts/components/Filter.js" defer="true"></script>
+    <script type="text/javascript" src="scripts/scaladoc-scalajs.js" defer="true"></script>
+    <script type="text/javascript" src="scripts/contributors.js" defer="true"></script>
+    <script type="text/javascript" src="https://code.jquery.com/jquery-3.5.1.min.js" defer="true"></script>
```
errt: Do we really need these pinned-version external scripts? They might unnoticeably become outdated.

errt (changes requested):

TOOLS/ce/src/main/resources/template.html
johannesduesing: I managed to remove almost all JS libraries from the template 👍

johannesduesing:
I can confirm that `buildAll` still generates the CE HTML file under `target/scala-3.7.3/unidoc/config.html`. This PR should be good to merge.

errt (approved):

PR #356, opened 2026-01-16
Remove broken it-test

errt:
Remove it-test that depends on Hermes being present in OPAL. Note that the test should test the Hermes implementation from the resource file, but doesn't since ClassLoaders first resolve using their parent class loader before trying to find a class themselves. With Hermes removed from OPAL, it does indeed invoke the Hermes implementation from the resource file but that uses `this.getClass.getResourceAsStream("HermesCLI.txt")` which does exist int he resource file, but is not found by the ClassLoader. This surely could be fixed, but the test doesn't seem important enough to jump through hoops.


johannesduesing (approved):
Agreed, it does not make sense to put any work into fixing this issue - lgtm 👍

PR #362, opened 2026-05-12
Fixes scala-reflect issue with GitHub pureconfig (https://github.com/opalj/JCG/issues/17)

svenkeidel:
Hi,

This fixes https://github.com/opalj/JCG/issues/17.

The issue is that Opal depends on scala-reflect, which is binary incompatible with Scala 3 classes.
The problem is that Githubs pureconfig generates Scala 3 code that crashes scala-reflect searches for `ConfigurationEntryPointsFinder`.

The fix reimplements the `ConfigurationEntryPointsFinder.EntryPointContainer` configuration reader to avoid the generation of problematic Scala 3 classes.

I tested the fix in JCG and it works.

Sven

### DIFF
--- a/OPAL/br/src/main/scala/org/opalj/br/analyses/cg/EntryPointFinder.scala
+++ b/OPAL/br/src/main/scala/org/opalj/br/analyses/cg/EntryPointFinder.scala
@@ -380,7 +380,8 @@ trait ConfigurationEntryPointsFinder extends EntryPointFinder {
 
     // Implement ConfigReader manually to avoid issue with scala-reflect: https://github.com/opalj/JCG/issues/17
     implicit private val entryPointContainerReader: ConfigReader[EntryPointContainer] =
-        ConfigReader.forProduct3("declaringClass", "name", "descriptor")(new EntryPointContainer(_,_,_))
+        ConfigReader.forProduct3("declaring-class", "name", "descriptor")(EntryPointContainer.apply)
+          .orElse(ConfigReader.forProduct3("declaringClass", "name", "descriptor")(EntryPointContainer.apply))
 }
 
 /**
--- a/OPAL/br/src/main/scala/org/opalj/br/analyses/cg/EntryPointFinder.scala
+++ b/OPAL/br/src/main/scala/org/opalj/br/analyses/cg/EntryPointFinder.scala
@@ -379,9 +379,8 @@ trait ConfigurationEntryPointsFinder extends EntryPointFinder {
     )
 
     // Implement ConfigReader manually to avoid issue with scala-reflect: https://github.com/opalj/JCG/issues/17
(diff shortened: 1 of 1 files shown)


OPAL/br/src/main/scala/org/opalj/br/analyses/cg/EntryPointFinder.scala
```diff
@@ -376,7 +376,12 @@ trait ConfigurationEntryPointsFinder extends EntryPointFinder {
         declaringClass: String,
         name:           String,
         descriptor:     Option[String]
-    ) derives ConfigReader
+    )
+
+    // Implement ConfigReader manually to avoid issue with scala-reflect: https://github.com/opalj/JCG/issues/17
+    implicit private val entryPointContainerReader: ConfigReader[EntryPointContainer] =
+        ConfigReader.forProduct3("declaring-class", "name", "descriptor")(EntryPointContainer.apply)
+          .orElse(ConfigReader.forProduct3("declaringClass", "name", "descriptor")(EntryPointContainer.apply))
```
errt: Why this? There should only be one way to write the keys that is valid now.

svenkeidel: - `reference.conf` uses "declaring-class":
https://github.com/opalj/opal/blob/55085cfc09d6ece415477c4ec206c4ecaab8d1dc/OPAL/br/src/main/resources/reference.conf#L96-L111

- `org.opalj.br.analyses.cg.InitialEntryPointsKey.entryPoints` uses "declaringClass": https://github.com/opalj/JCG/blob/master/jcg_opal_testadapter/src/main/scala/OpalJCGAdatper.scala#L82-L83

errt: The latter is because JCG has not been updated to recent Scala-3 based OPAL which, due to changes in the config library, had to change to "declaring-class". I.e., I think JCG has to be changed here instead of supporting different keys in OPAL.

svenkeidel: Documentation uses "declaringClass" too: https://github.com/opalj/opal/blob/55085cfc09d6ece415477c4ec206c4ecaab8d1dc/OPAL/br/src/main/scala/org/opalj/br/analyses/cg/EntryPointFinder.scala#L233-L243

Probably should be updated as well.

errt: Absolutely. Want to do that in this PR?

svenkeidel:
Done.

errt (approved):
LGTM

PR #363, opened 2026-05-13
Migrate from scala-reflect 2.13 to Java's reflection API

svenkeidel:
Motivated by several issues with scala-reflect 2.13, this PR migrates to Java's reflection API.

This compiles, but still needs to be tested thoroughly.


errt (approved):
Looking good, thanks!

Issue #357, opened 2026-02-17
Fix Usage of ReflectionAllocations Analysis


johannesduesing:
As discussed in #353, we currently see error output during tests that suggests the `PointsToCallGraphKey` tries to load the configured `ReflectionAllocationsAnalysisScheduler` and fails doing so, since it appends and prepends text to obtain the total name: `org.opalj.tac.fpcf.analyses.pointsto.TypeBasedReflectionAllocationsAnalysisSchedulerAnalysisScheduler`. This class obviously does not exist, therefore loading fails. There seem to be multiple instance of this error.

However, the test cases are still pass - this is likely a bug. On top of that, that particular analysis may not always be required. This should be adapted accordingly.

PR #366, opened 2026-07-02
Fix duplicate update in callees

johannesduesing:
This PR fixes a duplicate, irrelevant call to `updateWith` for the direct callees in `Callees.scala`. I tried to test the performance improvement resulting from this, but found no measurable impact when running the `CallGraphIntegrationTest` multiple times. I did, however, see errors when loading the *TypeBased* `ReflectionAllocationsAnalysisScheduler` - an additional `AnalysisScheduler` was appended to the class name before reflectively locating the class, which resulted in a ClassNotFoundException. Using the fully qualified name in the configuration, the error disappeared.


errt:
The latter is a known issue: https://github.com/opalj/opal/issues/357

johannesduesing:
As discussed in person, i've reverted the configuration change to be addressed in a separate issue.

errt (approved):

